PR #30, opened 2020-02-14
Update api config

andrewxhill:


src/ThreadsConfig.ts
```diff
@@ -13,14 +13,15 @@ export class ThreadsConfig extends Config {
...
+    public scheme: string = 'https',
+    public authApi: string = 'cloud.textile.io',
+    public authPort: number = 443,
+    public threadsApi: string = 'api.textile.io',
     public threadsPort: number = 6007,
     public transport: grpc.TransportFactory = grpc.WebsocketTransport(),
   ) {
     super()
-    this.host = `${this.apiScheme}://${this.api}:${this.threadsPort}`
+    this.host = `${this.scheme}://${this.threadsApi}:${this.threadsPort}`
```
carsonfarmer: Might be nice to have a catch here as to whether port is undefined or not. Then only add the port if !== undefined?

carsonfarmer: Something like:
```suggestion
    this.host = `${this.scheme}://${this.threadsApi}${this.threadsPort && ':' + this.threadsPort}`
```

andrewxhill: cool, we can do that as an enhancement if ever requested

PR #34, opened 2020-02-19
config: easy transport switch for node apps

andrewxhill:
Signed-off-by: andrew <andrew@textile.io>


carsonfarmer:
I'm not 100% sure we want this here. I'd like to think a bit more about the best place to configure this for devs working on nodejs.

carsonfarmer:
This is no longer needed as we've moved to a 'Context' based API. Please reopen if needed.

PR #75, opened 2020-04-01
First docs build

andrewxhill:


andrewxhill:
fixed the hierarchy a bit, so it's not API=>API. now it's,

Screen Shot 2020-04-01 at 3 48 25 PM

carsonfarmer:
This is great! It also matches how js-threads sub-modules all show up in docs as well, which is slick!

PR #85, opened 2020-04-13
threads: update to new api with newStore=>newDB

andrewxhill:


package.json
```diff
@@ -57,7 +57,8 @@
   "license": "MIT",
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.12.0",
-    "@textile/threads-client": "^0.2.6",
+    "@textile/threads-client": "^0.2.8",
+    "@textile/threads-core": "0.0.3-alpha.3",
```
andrewxhill, line 61: making ThreadID available here easy

package.json
```diff
@@ -57,7 +57,8 @@
   "license": "MIT",
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.12.0",
-    "@textile/threads-client": "^0.2.6",
+    "@textile/threads-client": "^0.2.8",
```
andrewxhill, line 60: this is the only real change, the rest is just tests/example reformatting to use the newDB api

PR #109, opened 2020-04-24
Adding hub apis + context/credentials for managing headers

carsonfarmer:


carsonfarmer:
@andrewxhill This adds all hub methods and full test coverage. It requires the _latest_ docker image to be running locally, based primarily on your instructions. I'll get that into the CI here next, so we can make sure it runs locally and passes tests within CI.

PR #117, opened 2020-04-28
Buckets and Users APIs

carsonfarmer:


carsonfarmer:
Ok, @andrewxhill this is ready for a review. I've now removed the hub client code, and I'm using the latest docker image which mimics @jsign's `local-up` setup. This currently exports Context and Buckets from the main target. I'll branch off this PR and add the User APIs shortly as a separate PR.

carsonfarmer:
Note that we'll have to update the @textile/hub-grpc and @textile/buckets-grpc packages in this package once we officially release those.

src/buckets.spec.ts
```diff
@@ -0,0 +1,200 @@
+/* eslint-disable import/first */
```
carsonfarmer, line 1: These tests are all pretty much exactly the same as the go versions. The only difference here is I only create one account and one bucket and keep using that one along the way. Note that there are tests specific to browser and node. This is a good flow to use for our other packages that need to run different code in different envs as well.

src/buckets.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  it('should push data from file API in browser', async function () {
+    if (isNode) return this.skip()
+    const parts = [
+      new Blob(['you construct a file...'], { type: 'text/plain' }),
+      ' Same way as you do with blob',
+      new Uint16Array([33]),
+    ]
+    // Construct a file
+    const file = new File(parts, 'file1.txt')
```
carsonfarmer, line 106: This is the equivalent of getting a file object from a drag and drop event or a file picking in the browser, which is likely a pretty standard use case. The nice thing here is we can essentially stream the file to our grpc method, but it feels like a standard promise-based rest call to the developer.

src/buckets.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  })
+
+  it('should pull files by path and write to file on node', async function () {
+    if (isBrowser) return this.skip()
+    // Bucket path
+    const rootKey = buck.root?.key || ''
+    let length = 0
+
+    // Bucket path
+    const chunks = client.pullPath(rootKey, 'dir1/file1.jpg', ctx, {
```
carsonfarmer, line 150: This returns an AsyncIterableIterator of UInt8Arrays (Buffers), which can easily be fed into a write stream, or file write object!

src/buckets.ts
```diff
@@ -0,0 +1,253 @@
...
+    })
+    return chan.wrap(() => response.close())
+  }
+
+  private unary<
+    R extends grpc.ProtobufMessage,
+    T extends grpc.ProtobufMessage,
+    M extends grpc.UnaryMethodDefinition<R, T>
+  >(methodDescriptor: M, req: R, context?: Context): Promise<T> {
+    // @todo: This is not totally ideal, but is cleaner for returning promises.
```
carsonfarmer, line 228: Ideally we'd just use the client directly, but that ends up creating a lot of boilerplate code. So I just do what we've done elsewhere here (but with slightly better type inference here).

src/buckets.ts
```diff
@@ -0,0 +1,253 @@
...
+    ctx?: Context,
+    opts?: { progress?: (num?: number) => void },
+  ): AsyncIterableIterator<Uint8Array> {
+    const metadata = this.context.withContext(ctx).toJSON()
+    const chan = new Channel<Uint8Array>()
+    const request = new pb.PullPathRequest()
+    request.setKey(key)
+    request.setPath(path)
+    let written = 0
+    const response = grpc.invoke(API.PullPath, {
```
carsonfarmer, line 199: This is shorthand for creating a client and calling the pullPath method. We _could_ also do it that way, but this felt cleaner.

src/buckets.ts
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Pushes a file to a bucket path.
+   * @param key Unique (IPNS compatible) identifier key for a bucket.
+   * @param path A file/object (sub)-path within a bucket.
+   * @param input The input file/stream/object.
+   * @param ctx Context object containing web-gRPC headers and settings.
+   * @param opts Options to control response stream. Currently only supports a progress function.
+   * @note This will return the resolved path and the bucket's new root path.
+   */
+  async pushPath(key: string, path: string, input: any, ctx?: Context, opts?: { progress?: (num?: number) => void }) {
```
carsonfarmer, line 116: If you follow this along, it matches the go version quite closely. The fun difference here is, we return a promise, which resolves with the final message from the server. So we're wrapping the whole bidi dance into a single promise. It feels like a nice abstraction for the developer, so callbacks (unless you want progress) needed.

src/buckets.ts
```diff
@@ -0,0 +1,253 @@
...
+   * @param opts Options to control response stream. Currently only supports a progress function.
+   */
+  pullPath(
+    key: string,
+    path: string,
+    ctx?: Context,
+    opts?: { progress?: (num?: number) => void },
+  ): AsyncIterableIterator<Uint8Array> {
+    const metadata = this.context.withContext(ctx).toJSON()
+    const chan = new Channel<Uint8Array>()
```
carsonfarmer, line 194: This is cool, its a writable AsyncIterableIterator (basically) with back pressure support! Behaves a lot like a go chan.

src/normalize.ts
```diff
@@ -0,0 +1,321 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
carsonfarmer, line 1: This is just a typescript version of some js-ipfs code!

src/context.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Org slug/name. Used for various org session operations.
+   */
+  ['x-textile-org']?: string
+}
+
+/**
+ * Context provides immutable context management for gRPC credentials and config settings.
+ */
+export class Context implements Config {
```
carsonfarmer, line 31: I love the go context pattern, so we've got something similar here. I think it'll be useful for a) specifying headers without having to know much about them, and b) mapping docs between js and go. Also, what a great design pattern! This version is quite opinionated in terms of which types of keys it'll accept (you have to extend it each time). Maybe in the future, I'll create a generic version of this. For now, this works great. Note also that it compiles with our existing Config interface, so we can move that stuff over more slowly.

src/buckets.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  let ctx: Context = new Context(addrApiurl, undefined)
+  const client = new Buckets(ctx)
+  let buck: InitReply.AsObject
+  let fileSize: number
+  before(async () => {
+    const user = await signUp(ctx, addrGatewayUrl, sessionSecret)
+    ctx = ctx.withSession(user.user?.session).withThreadName('buckets')
+    const id = ThreadID.fromRandom()
+    const db = new Client(ctx)
+    // @todo: Warning, this is a hack!
```
carsonfarmer, line 31: @andrewxhill we'll have to update js-threads-client to accept a Context here, otherwise, it just ignores all the extra headers and only uses host and transport. Easy fix, which once done, we can update these tests.

andrewxhill: sgtm

src/context.ts
andrewxhill: thinking through this. since `Context` implements `Config`, doesn't threads-client already accept `Context` (since it accepts `Config`)?  I guess you meant above that we need to include a few more of these headers. 

One rub here... I can't import Context in threads-client (since it'll be circular dep). So perhaps we split Context out of here?

andrewxhill: actually... it looks like i don't need to implement anything new from here since you are using the wrap methods primarily. so it _should_ just be gtg right?

src/context.ts
```diff
@@ -0,0 +1,109 @@
+import { grpc } from '@improbable-eng/grpc-web'
+import { ThreadID } from '@textile/threads-id'
+import { Config } from '@textile/threads-client'
+
+type HostString = 'https://hub.textile.io:443' | 'https://hub.staging.textile.io:443' | 'http://127.0.0.1:3007' | string
```
andrewxhill, line 5: since it can just be a `string` what is the point of the other options that are just strings?

src/buckets.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+const wrongError = new Error('wrong error!')
+const sessionSecret = 'textilesession'
+
+describe('Buckets...', () => {
+  let ctx: Context = new Context(addrApiurl, undefined)
+  const client = new Buckets(ctx)
+  let buck: InitReply.AsObject
+  let fileSize: number
+  before(async () => {
+    const user = await signUp(ctx, addrGatewayUrl, sessionSecret)
```
andrewxhill, line 27: just walking through the js todo example wiring this up. 

how should another app use this? should i import `pb` and use the `pb.signup..` method directly in the todo app or is there another way?

src/context.ts
carsonfarmer: Correct

carsonfarmer: The only sticking point is the 'hack' in the comment below. Since js-threads-client checks that the input isinstance of Config (and Context isn't) it basically 'wipes' out all the internal headers. Hence the 'hack' to patch it back on below. So if we adjust that check, we'll be just fine.

src/context.ts
carsonfarmer: Autocomplete :)

carsonfarmer: (also mostly just reminders for myself)

src/buckets.spec.ts
carsonfarmer: No, you shouldn't use this at all. You should do this part all from the cli.

src/users.spec.ts
```diff
@@ -103,12 +104,97 @@ describe('Users...', () => {
...
+      const id = ThreadID.fromRandom()
+      const db = new Client(ctx)
+      await db.newDB(id.toBytes())
+      res = await client.listThreads(ctx)
+      expect(res.listList).to.have.length(1)
+    })
+
+    it('should handle users keys', async () => {
+      const key = await createKey(ctx.withSession(dev.session), 'USER')
+      const sig = await createAPISig(key.secret) // Defaults to 1 minute from now
```
andrewxhill, line 171: expiration in 1 minute?

src/buckets.ts
sanderpick: Very nice

src/context.ts
sanderpick: You shouldn't need the port on the hub ones, right? I can't remember, is the proxy on 447?

src/context.ts
```diff
@@ -0,0 +1,152 @@
...
+export const defaultHost: HostString = 'https://hub.textile.io:443'
+
+export interface ContextKeys {
```
sanderpick, line 8: This is great!

src/users.ts
```diff
@@ -0,0 +1,68 @@
...
+    })
+  }
+
+  /**
+   * Returns a Thread by name.
+   * @param name The name of the Thread.
+   * @param ctx Context containing gRPC headers and settings.
+   * These will be merged with any internal credentials.
+   */
+  async getThread(name: string, ctx?: Context) {
```
sanderpick, line 32: 🎉

PR #151, opened 2020-05-12
chore(deps): adds external context dep

andrewxhill:


andrewxhill:
okay, first error

```
  1) Users...
       getThread
         should handle bad keys:

      AssertionError: expected 5 to equal 16
      + expected - actual

      -5
      +16
      
      at /home/runner/work/js-textile/js-textile/src/users.spec.ts:51:29
      at Generator.throw (<anonymous>)
      at rejected (src/users.spec.ts:6:65)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

The actual error coming back is 'Thread not found'. which is strange since you'd think auth would come before exists

andrewxhill:
The second issue

```
  2) Users...
       getThread
         should handle users keys:

```

is erroring out because it's not throwing an error when the token doesn't exist

carsonfarmer:
Hmmm, so now that Context is mutable, its possible that we _are_ properly authenticated, and it needs to be cleared out?

andrewxhill:
Okay, same errors with 3 and 4. will try to scope the context a bit more

andrewxhill:
what do you make of error 1 though?

carsonfarmer:
taking a look now

carsonfarmer:
The issue was indeed that the Context was being mutated underneath us due to the fact that it is now mutable. The mutable API is easier to hide, but harder to deal with. Its a trade off I guess? Anyway, I have also tried to reduce the number of times we actually reference the context in the tests to hopefully test how well we can 'hide' it. Also, this unexpected mutation is really only an issue for tests, where we _want_ to throw errors etc.

PR #154, opened 2020-05-12
Fully isolate user context

carsonfarmer:
@andrewxhill , I've updated this test to ensure that the user context is _fully isolated_ from the developer context... that way, we can be sure this is a realistic usage scenario.

### DIFF
--- a/src/users.spec.ts
+++ b/src/users.spec.ts
@@ -238,6 +238,7 @@ describe('Users...', () => {
     context('a developer with a user', () => {
       const ctx: Context = new Context(addrApiurl, undefined)
       let dev: SignupReply.AsObject
+      const userContext = new Context(addrApiurl)
       it('should sign-up, create an API key, and sign it for the requests', async () => {
         // @note This should be done using the cli
         const { user } = await signUp(ctx, addrGatewayUrl, sessionSecret)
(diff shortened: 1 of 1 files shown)


src/users.spec.ts
```diff
@@ -246,24 +247,26 @@ describe('Users...', () => {
         // @note This should be done using the cli
         // This time they create a user key
         const key = await createKey(ctx, 'USER')
-        await ctx.withAPIKey(key.key).withUserKey(key)
-        expect(ctx.toJSON()).to.have.ownProperty('x-textile-api-sig')
+        // Update user context
+        await userContext.withAPIKey(key.key).withUserKey(key)
```
carsonfarmer, line 251: Fully isolated user context

src/users.spec.ts
```diff
@@ -246,24 +247,26 @@ describe('Users...', () => {
...
+        const db = new Client(userContext)
         await db.getToken(identity)
       })
       it('should then create a db for the bucket', async () => {
         const id = ThreadID.fromRandom()
-        const db = new Client(ctx)
-        await db.newDB(id, ctx.withThreadName('my-buckets'))
-        expect(ctx.toJSON()).to.have.ownProperty('x-textile-thread-name')
+        // Just for the test, recrete the client here
+        const db = new Client(userContext)
```
carsonfarmer, line 262: Also doing this to demonstrate that, in an app that refreshes on a page load, this would still work if you (say) stored the context in redux or similar.

PR #157, opened 2020-05-12
Proposed simplification of users APIs

carsonfarmer:
@andrewxhill, here's a proposed simplified workflow for you. I'll make comments inline, please let me know what you think.

### DIFF
--- a/src/buckets.ts
+++ b/src/buckets.ts
@@ -6,7 +6,6 @@ import { Channel } from 'queueable'
 import { grpc } from '@improbable-eng/grpc-web'
 import { Context } from '@textile/context'
 import { normaliseInput, File } from './normalize'
-import { Users } from './users'
 
 const logger = log.getLogger('buckets')
 
@@ -36,10 +35,6 @@ export class Buckets {
       transport: context.transport,
       debug: context.debug,
     }
-  }
-
-  static fromUser(user: Users) {
-    return new Buckets(user.context)
   }
 
   /**
@@ -230,8 +225,6 @@ export class Buckets {
     T extends grpc.ProtobufMessage,
     M extends grpc.UnaryMethodDefinition<R, T>
   >(methodDescriptor: M, req: R, context?: Context): Promise<T> {
-    // @todo: This is not totally ideal, but is cleaner for returning promises.
-    // Ideally, we'd use the generated client directly, and wrap that with a promise.
     return new Promise<T>((resolve, reject) => {
       const creds = this.context.withContext(context)
       grpc.unary(methodDescriptor, {
--- a/src/buckets.ts
+++ b/src/buckets.ts
@@ -6,6 +6,7 @@ import { Channel } from 'queueable'
 import { grpc } from '@improbable-eng/grpc-web'
 import { Context } from '@textile/context'
 import { normaliseInput, File } from './normalize'
+import { Users } from './users'
 
 const logger = log.getLogger('buckets')
 
@@ -35,6 +36,10 @@ export class Buckets {
       transport: context.transport,
       debug: context.debug,
     }
+  }
+
+  static fromUser(user: Users) {
+    return new Buckets(user.context)
   }
 
   /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
 export { Context } from '@textile/context'
 export * from './buckets'
 export * from './users'
-export { createAPISig } from './utils'
+export { Client } from '@textile/threads-client'
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,3 @@
 export { Context } from '@textile/context'
 export * from './buckets'
 export * from './users'
-export { Client } from '@textile/threads-client'
--- a/src/users.spec.ts
+++ b/src/users.spec.ts
@@ -235,38 +235,37 @@ describe('Users...', () => {
         expect(res.id).to.deep.equal(ctx.toJSON()['x-textile-thread'])
       })
     })
-    context('a developer with a user', () => {
+    context('a developer with a user', function () {
       const ctx: Context = new Context(addrApiurl, undefined)
       let dev: SignupReply.AsObject
-      const userContext = new Context(addrApiurl)
-      it('should sign-up, create an API key, and sign it for the requests', async () => {
+      let users: Users
+      it('should sign-up, create an API key, and a new user', async function () {
         // @note This should be done using the cli
         const { user } = await signUp(ctx, addrGatewayUrl, sessionSecret)
         if (user) dev = user
-        ctx.withSession(dev.session)
         // @note This should be done using the cli
         // This time they create a user key
-        const key = await createKey(ctx, 'USER')
-        // Update user context
-        await userContext.withAPIKey(key.key).withUserKey(key)
-        expect(userContext.toJSON()).to.have.ownProperty('x-textile-api-sig')
-      }).timeout(3000)
-      it('should then generate a user identity and get a token for it', async () => {
+        const key = await createKey(ctx.withSession(dev.session), 'USER')
+
+        // This should automatically generate a user identity and validate keys, though we use a random ident
+        // for demo purposes here to show that it can also use custom identities
         const identity = await Libp2pCryptoIdentity.fromRandom()
-        const db = new Client(userContext)
-        await db.getToken(identity)
+        // We also explicitly specify a custom context here, which could be omitted as it uses reasonable defaults
+        const userContext = new Context(addrApiurl)
+        // In the app, we simply create a new user from the provided user key, signing is done automatically
+        users = await Users.fromKey(key, identity, userContext)
+        expect(users.context.toJSON()).to.have.ownProperty('x-textile-api-sig')
+      }).timeout(3000)
+
+      it('should then create a db for the bucket', async function () {
+        await users.createThread('my-buckets')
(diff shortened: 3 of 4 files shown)


src/buckets.ts
```diff
@@ -37,6 +38,10 @@ export class Buckets {
     }
   }
 
+  static fromUser(user: Users) {
```
carsonfarmer, line 41: Just some syntactic sugar to make it easy to create a bucket api instance from an existing user instance

src/index.ts
```diff
@@ -1,4 +1,4 @@
 export { Context } from '@textile/context'
 export * from './buckets'
 export * from './users'
-export { createAPISig } from './utils'
+export { Client } from '@textile/threads-client'
```
carsonfarmer, line 4: Its already a dep, and going to be used 99% of the time

src/users.spec.ts
```diff
@@ -235,38 +235,37 @@ describe('Users...', () => {
...
-    context('a developer with a user', () => {
+    context('a developer with a user', function () {
       const ctx: Context = new Context(addrApiurl, undefined)
       let dev: SignupReply.AsObject
-      const userContext = new Context(addrApiurl)
-      it('should sign-up, create an API key, and sign it for the requests', async () => {
+      let users: Users
+      it('should sign-up, create an API key, and a new user', async function () {
         // @note This should be done using the cli
         const { user } = await signUp(ctx, addrGatewayUrl, sessionSecret)
```
carsonfarmer, line 244: So the new workflow becomes:
1. Via the CLI or eventual dashboard...
  a. Dev signs up etc
  b. They generate a user key
  c. They find some way to pass this key info to the app
2. Then in the app
  a. Create an identity (we use a random one by default)
  b. Create a new user from a key and an identity (with optional context if they are advanced)
  c. Create a thread for a bucket if they want (or even for a db)
  d. Create a bucket from a user, and you're good to go.

carsonfarmer: All of the above are optional except the key bit. The signing etc is done under the hood. So in theory, this is a one-liner once they have a key and can serve it up to their app.

src/users.ts
```diff
@@ -50,6 +62,19 @@ export class Users {
     })
   }
 
+  /**
+   * Create a new Thread with name.
+   * @param name The name of the Thread.
+   * @param ctx Context containing gRPC headers and settings.
+   * These will be merged with any internal credentials.
+   */
+  async createThread(name: string, ctx?: Context) {
```
sanderpick, line 71: Poking my nose in here :) 

I don't think I am understanding something here. Can't the "user" just leverage the threads client directly?

src/users.spec.ts
```diff
@@ -235,38 +235,37 @@ describe('Users...', () => {
...
-        const id = ThreadID.fromRandom()
-        // Just for the test, recrete the client here
-        const db = new Client(userContext)
-        await db.newDB(id, userContext.withThreadName('my-buckets'))
-        expect(userContext.toJSON()).to.have.ownProperty('x-textile-thread-name')
+        // We also explicitly specify a custom context here, which could be omitted as it uses reasonable defaults
+        const userContext = new Context(addrApiurl)
+        // In the app, we simply create a new user from the provided user key, signing is done automatically
+        users = await Users.fromKey(key, identity, userContext)
+        expect(users.context.toJSON()).to.have.ownProperty('x-textile-api-sig')
```
carsonfarmer, line 257: I'm digging into the context here just for the tests

src/users.ts
```diff
@@ -50,6 +62,19 @@ export class Users {
     })
   }
 
+  /**
+   * Create a new Thread with name.
```
carsonfarmer, line 66: This is just a nice to have

src/users.ts
carsonfarmer: Yes they can, this is just syntactic sugar for a common workflow

sanderpick: Ok, but, two things: 
- You don't think it's confusing to have the same method in two places? 
- Looks like we're creating a DB here, so wouldn't it be `newDB`?

carsonfarmer: I don't think it'll be confusing, because its possible they won't _ever_ actually touch the threads client at all if they're only ever using the user and buckets apis. This allows them to create a thread with less work (we had it the other way before and that's what has lead to some confusion it seems). To your second point... yup good point!

carsonfarmer: I could certainly be convinced to revert this (as that was the original plan) but I'm trying to come up with _some_ way to make it so a developer who is "just trying to create a bucket for a user" doesn't have to think about databases etc.

carsonfarmer: Hopefully that 'motivation' makes sense?

carsonfarmer: At some point, that could defs make sense, yes. Right now, we're just trying to find a simple workflow that doesn't require too many steps to get started on the hub. I kind of 'like' being able to scope things out like this, so `import { Users }` to to do user things, `import { Buckets }` to do buckets things, and `import { Client }` to do threads things... with the ability to pass the same Context between all of these.

src/buckets.ts
carsonfarmer: This might actually be a bit much now that I think about it, just as easy to do `new Bucket(user.context)`

carsonfarmer: Will revert

src/users.ts
sanderpick: Okay, sounds good. I think it might not be clear, or just plain confusing, what the role of the User API is. It likely only makes sense as a low level thing: The only thing you can't do with the threads and buckets clients is _list all threads created with your API key or Account session_ (since those clients are not multi-tenant by nature) which gave birth to this Users API thing. So, hiding it behind a higher level client feels smart... maybe there's just Buckets and Threads? Then, behind the scenes, the _two_ Users API methods are tacked onto Threads?

sanderpick: So, two options: 
1. Alias one of the threads methods onto users (what you've done here)
2. Alias the two users methods onto threads (which would kill one of the app facing clients, which sounds like a good thing)

carsonfarmer: I like it, let's do that (actually, now that you mention it, that was the original plan from @andrewxhill )

sanderpick: Sweet, sounds good 🎸

src/users.spec.ts
carsonfarmer: Ok, with updates this is slightly more verbose, but cleaner thanks to some discussion with @sanderpick

src/users.spec.ts
```diff
@@ -230,43 +229,43 @@ describe('Users...', () => {
...
+        // We also explicitly specify a custom context here, which could be omitted as it uses reasonable defaults
+        const userContext = await new Context(addrApiurl).withUserKey(key)
+        // In the app, we simply create a new user from the provided user key, signing is done automatically
+        users = new Client(userContext)
+        await users.getToken(identity)
+        expect(users.context.toJSON()).to.have.ownProperty('x-textile-api-sig')
+      }).timeout(3000)
+
+      it('should then create a db for the bucket', async function () {
+        await users.newDB(ThreadID.fromRandom(), users.context.withThreadName('my-buckets'))
```
carsonfarmer, line 261: This could be cleaned up slightly

src/users.ts
```diff
@@ -4,76 +4,77 @@ import { APIClient } from '@textile/users-grpc/users_pb_service'
...
-  /**
-   * Creates a new gRPC client instance for accessing the Textile Buckets API.
-   * @param context The context to use for interacting with the APIs. Can be modified later.
-   */
-  constructor(public context: Context = new Context()) {
-    this.client = new APIClient(context.host, {
-      transport: context.transport,
-      debug: context.debug,
-    })
+// Module augmentation to add methods to Peer from core
```
carsonfarmer, line 11: This is pretty cool, we essentially extend the Client interface on the fly here, and then the implementations follow

src/users.ts
```diff
@@ -4,76 +4,77 @@ import { APIClient } from '@textile/users-grpc/users_pb_service'
...
-  }
+/**
+ * Returns a Thread by name.
+ * @param name The name of the Thread.
+ * @param ctx Context containing gRPC headers and settings.
+ * These will be merged with any internal credentials.
+ */
+Client.prototype.getThread = async function (name: string, ctx?: Context) {
+  logger.debug('get thread request')
+  const client = new APIClient(this.serviceHost, {
```
carsonfarmer, line 27: Because we're 'piggy backing' off of threads client, we create out own client here each time, but use all the options and settings from threads client.

src/users.ts
```diff
@@ -4,76 +4,77 @@ import { APIClient } from '@textile/users-grpc/users_pb_service'
...
+            thread.id = ThreadID.fromBytes(Buffer.from(thread.id as string, 'base64')).toString()
+          })
+        }
+        resolve(msg)
+      },
+    )
+  })
 }
+
+export { Client }
```
carsonfarmer, line 80: Exported for convinience, but these new methods should exist even if we import and let side-effects take over

carsonfarmer:
With these changes, if you do:

```typescript
import { Client, Buckets } from '@textile/textile`
```

you essentially have all the hub-based APIs you'd want. This PR augments Client with User API methods getThread and listThreads, so developers can more easily manage threads without having to interact with a whole new set of APIs, or the low level Network Client APIs.

If they aren't using Buckets, they could also do:

```typescript
import '@textile/textile` and still get the augmented Client with the Users APIs, as this code also provides side-effects to augment the Client.
```

PR #176, opened 2020-05-20
Switch to lerna structure

carsonfarmer:


package.json
```diff
@@ -5,23 +5,16 @@
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
```
andrewxhill, line 6: I noticed that in js-textile, this was "dist/index". any reason for the difference?

carsonfarmer: yeh we should leave off the suffix

carsonfarmer: But they aren't used here anyway, so will just remove them

PR #177, opened 2020-05-20
feat(buckets): adds standalone buckets lib

andrewxhill:


packages/buckets/src/buckets.ts
```diff
@@ -4,11 +4,18 @@ import { API, APIPushPath } from '@textile/buckets-grpc/buckets_pb_service'
 import CID from 'cids'
 import { Channel } from 'queueable'
 import { grpc } from '@improbable-eng/grpc-web'
-import { Context } from '@textile/context'
+import { Context as Ctx } from '@textile/context'
```
carsonfarmer, line 7: So here we still import Context lib, but alias it to Ctx because we're only using it to create a default Context. I'm calling this a 'minimal' fix because, in typescript, if we reference buckets.context, it'll only satisfy the minimal Context interface defined here, which isn't going to work _outside_ Buckets (it doesn't have all the with* methods. Ideally, we'd export the full interface from @textile/context and use that here, but to keep things simple an to avoid having to make those changes right away, we can use this minimal fix.

packages/buckets/src/buckets.ts
```diff
@@ -4,11 +4,18 @@ import { API, APIPushPath } from '@textile/buckets-grpc/buckets_pb_service'
...
 import { normaliseInput, File } from './normalize'
 
 const logger = log.getLogger('buckets')
 
+export interface Context {
```
carsonfarmer, line 12: So this is the minimal interface we need to work in Buckets.

packages/buckets/src/buckets.ts
```diff
@@ -203,7 +211,7 @@ export class Buckets {
     ctx?: Context,
     opts?: { progress?: (num?: number) => void },
   ): AsyncIterableIterator<Uint8Array> {
-    const metadata = this.context.withContext(ctx).toJSON()
+    const metadata = { ...this.context.toJSON(), ...ctx?.toJSON() }
```
carsonfarmer, line 214: Cleaner, and doesn't require a withContext method on our interface

packages/buckets/src/buckets.ts
```diff
@@ -237,15 +245,15 @@ export class Buckets {
     R extends grpc.ProtobufMessage,
     T extends grpc.ProtobufMessage,
     M extends grpc.UnaryMethodDefinition<R, T>
-  >(methodDescriptor: M, req: R, context?: Context): Promise<T> {
+  >(methodDescriptor: M, req: R, ctx?: Context): Promise<T> {
     return new Promise<T>((resolve, reject) => {
-      const creds = this.context.withContext(context)
+      const metadata = { ...this.context.toJSON(), ...ctx?.toJSON() }
```
carsonfarmer, line 250: Same here

tsconfig.webpack.json
```diff
@@ -6,6 +6,7 @@
   },
   "exclude": [
     "**/*.spec.ts",
+    "**/*.util.ts",
```
carsonfarmer, line 9: nice

typedoc.json
```diff
@@ -8,7 +8,7 @@
     "dom",
     "esnext"
   ],
-  "exclude": ["**/*.spec.ts", "examples/**", "**/node_modules/**"],
+  "exclude": ["**/*.spec.ts", "**/spec.util.ts", "examples/**", "**/node_modules/**"],
```
carsonfarmer, line 11: smart!

andrewxhill: needed to get those deps over to devdeps

PR #214, opened 2020-06-04
chore(deps): upgrade @textile packages

andrewxhill:


.github/workflows/test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           TXTL_ADDR_GATEWAY_HOST: /ip4/0.0.0.0/tcp/8006
           TXTL_ADDR_MONGO_URI: mongodb://mongo:27017
           TXTL_ADDR_IPFS_API: /dns4/ipfs/tcp/5001
-          TXTL_EMAIL_SESSION_SECRET: textilesession
+          TXTL_EMAIL_SESSION_SECRET: hubsession
```
carsonfarmer, line 26: This is what the hub up make script now uses.

packages/buckets/src/buckets.spec.ts
```diff
@@ -60,7 +61,7 @@ describe('Buckets...', () => {
     expect(res.root).to.not.be.undefined
     expect(res.item?.isdir).to.be.true
     // @todo: Should we rename itemsList to just be items?
-    expect(res.item?.itemsList).to.have.length(0)
+    expect(res.item?.itemsList).to.have.length(1) // Includes .textileseed
```
carsonfarmer, line 64: These returns now include the .textileseed value, so we have to bump these all up by one

packages/textile/src/users.spec.ts
```diff
@@ -18,7 +18,7 @@ import { Client } from './users'
 const addrApiurl = 'http://127.0.0.1:3007'
 const addrGatewayUrl = 'http://127.0.0.1:8006'
 const wrongError = new Error('wrong error!')
-const sessionSecret = 'textilesession'
+const sessionSecret = 'hubsession'
```
carsonfarmer, line 21: updated in ci and via local makefile

packages/textile/src/users.spec.ts
```diff
@@ -72,10 +72,12 @@ describe('Users...', () => {
         expect(err.code).to.equal(grpc.Code.NotFound)
       }
       // All good
+      ctx.withThreadName('foo')
```
carsonfarmer, line 75: once https://github.com/textileio/js-threads/pull/263 is released, we can drop this

packages/textile/src/users.spec.ts
```diff
@@ -72,10 +72,12 @@ describe('Users...', () => {
...
       const id = ThreadID.fromRandom()
       const db = new Client(ctx)
-      await db.newDB(id, ctx.withThreadName('foo'))
-      const res = await client.getThread('foo')
+      await db.newDB(id, 'foo')
+      ctx.withThread(id.toString())
```
carsonfarmer, line 79: once https://github.com/textileio/js-threads/pull/263 this can also be dropped

packages/textile/src/users.spec.ts
```diff
@@ -72,10 +72,12 @@ describe('Users...', () => {
...
+      const res = await client.getThread('foo', ctx)
```
carsonfarmer, line 80: can also drop ctx from here

packages/textile/src/users.spec.ts
```diff
@@ -107,7 +109,8 @@ describe('Users...', () => {
       }
       // All good
       const id = ThreadID.fromRandom()
-      await db.newDB(id, ctx.withThreadName('foo'))
+      db.context.withThreadName('foo')
```
carsonfarmer, line 112: once https://github.com/textileio/js-threads/pull/263 lands, we can drop this

packages/textile/src/users.spec.ts
```diff
@@ -189,7 +192,8 @@ describe('Users...', () => {
       expect(res.listList).to.have.length(0)
       // Got one
       const id = ThreadID.fromRandom()
-      await db.newDB(id, ctx.withThreadName('foo'))
+      ctx.withThreadName('foo')
```
carsonfarmer, line 195: again, https://github.com/textileio/js-threads/pull/263 will fix this

packages/textile/src/users.spec.ts
```diff
@@ -211,9 +215,10 @@ describe('Users...', () => {
         expect(ctx.toJSON()).to.have.ownProperty('x-textile-api-sig')
       }).timeout(3000)
       it('should then create a db for the bucket', async () => {
+        ctx.withThreadName('my-buckets')
```
carsonfarmer, line 218: you get the pattern here

packages/textile/src/users.spec.ts
```diff
@@ -231,6 +236,7 @@ describe('Users...', () => {
         expect(root).to.not.be.undefined
 
         // We should have a thread named "my-buckets"
+        ctx.withThreadName('my-buckets')
```
carsonfarmer, line 239: also here

packages/textile/src/users.spec.ts
```diff
@@ -260,7 +266,9 @@ describe('Users...', () => {
       }).timeout(3000)
 
       it('should then create a db for the bucket', async function () {
-        await users.newDB(ThreadID.fromRandom(), users.context.withThreadName('my-buckets'))
+        // @todo https://github.com/textileio/js-threads/pull/263 should fix this...
+        users.context.withThreadName('my-buckets')
```
carsonfarmer, line 270: its in the todo

packages/textile/src/users.spec.ts
andrewxhill: hmm... that actually doesn't seem to be the case. i've done the update. everything else is now passing. if i remove this line it fails

packages/textile/src/users.spec.ts
andrewxhill: same. if i remove both it fails

packages/textile/src/users.spec.ts
andrewxhill: drop all three. still fails

packages/textile/src/users.spec.ts
andrewxhill: fails

packages/textile/src/users.spec.ts
andrewxhill: oop, looks like maybe i have a partial dep update. let me try that again.

andrewxhill: ooookay, pilot error. back on track

.github/workflows/test.yml
```diff
@@ -15,15 +15,16 @@ jobs:
 
     services:
       textile:
-        image: textile/textile:1.0.3
+        image: textile/textile:latest
```
carsonfarmer, line 18: living dangerously 😉

.github/workflows/test.yml
```diff
@@ -15,15 +15,16 @@ jobs:
...
         env:
-          TXTL_ADDR_API: /ip4/0.0.0.0/tcp/3006
-          TXTL_ADDR_API_PROXY: /ip4/0.0.0.0/tcp/3007
-          TXTL_ADDR_THREADS_HOST: /ip4/0.0.0.0/tcp/4006
-          TXTL_ADDR_GATEWAY_HOST: /ip4/0.0.0.0/tcp/8006
-          TXTL_ADDR_MONGO_URI: mongodb://mongo:27017
-          TXTL_ADDR_IPFS_API: /dns4/ipfs/tcp/5001
-          TXTL_EMAIL_SESSION_SECRET: textilesession
+          HUB_ADDR_API: /ip4/0.0.0.0/tcp/3006
```
carsonfarmer, line 20: ah good catch

packages/hub/src/users.spec.ts
```diff
@@ -18,36 +18,43 @@ import { Client } from './users'
...
+       */
       const tmp = new Context(addrApiurl).withSession(dev.session)
       const key = await createKey(tmp, 'ACCOUNT')
       try {
         await client.getThread('foo', ctx.withAPIKey(key.key))
         throw wrongError
       } catch (err) {
         expect(err).to.not.equal(wrongError)
-        expect(err.code).to.equal(grpc.Code.Unauthenticated)
+        expect(err.code).to.equal(grpc.Code.NotFound)
```
carsonfarmer, line 57: are we hoping that will be reverted to NotFound in a subsequent textile release?

packages/hub/src/users.spec.ts
```diff
@@ -74,7 +81,7 @@ describe('Users...', () => {
       // All good
       const id = ThreadID.fromRandom()
       const db = new Client(ctx)
-      await db.newDB(id, ctx.withThreadName('foo'))
+      await db.newDB(id, 'foo')
```
carsonfarmer, line 84: nice!

packages/hub/src/users.spec.ts
```diff
@@ -131,17 +138,18 @@ describe('Users...', () => {
...
+       * No key signature will pass because default security is null
+       */
       const tmp = new Context(addrApiurl).withSession(dev.session)
       const key = await createKey(tmp, 'ACCOUNT')
       try {
         await client.listThreads(ctx.withAPIKey(key.key))
         throw wrongError
       } catch (err) {
-        expect(err).to.not.equal(wrongError)
-        expect(err.code).to.equal(grpc.Code.Unauthenticated)
```
carsonfarmer, line 142: same here, should this be unauthenticated sometime in the future?

packages/hub/src/users.spec.ts
```diff
@@ -260,7 +269,9 @@ describe('Users...', () => {
       }).timeout(3000)
 
       it('should then create a db for the bucket', async function () {
-        await users.newDB(ThreadID.fromRandom(), users.context.withThreadName('my-buckets'))
+        // @todo https://github.com/textileio/js-threads/pull/263 should fix this...
+        users.context.withThreadName('my-buckets')
```
carsonfarmer, line 273:
```suggestion
```

carsonfarmer: I think this should be ok now right?

.github/workflows/test.yml
andrewxhill: yep, wont affect @textile/textile which is locked on the last release... forever.

packages/hub/src/users.spec.ts
andrewxhill: This is the current case. We want to be reverted to Unauthenticated

packages/hub/src/users.spec.ts
andrewxhill: yep. hopefully will update

PR #227, opened 2020-06-09
chore(docs): migrate to api-documenter and include more docs

andrewxhill:


.gitignore
```diff
@@ -28,6 +29,7 @@ node_modules/
 .pnp.js
 
 # Build Outputs
+markdown
```
andrewxhill, line 32: intermediate docs markdown will build here.

config/api-extractor.json
```diff
@@ -0,0 +1,369 @@
...
+   * analyzes the symbols exported by this module.
+   *
+   * The file extension must be ".d.ts" and not ".ts".
+   *
+   * The path is resolved relative to the folder of the config file that contains the setting; to change this,
+   * prepend a folder token such as "<projectFolder>".
+   *
+   * SUPPORTED TOKENS: <projectFolder>, <packageName>, <unscopedPackageName>
+   */
+  "mainEntryPointFilePath": "../packages/hub/dist/index.d.ts",
```
andrewxhill, line 48: docs come from the compiled lib, which is pretty nice

config/api-extractor.json
```diff
@@ -0,0 +1,369 @@
...
+   * and another NPM package "library2" is embedded in this bundle.  Some types from library2 may become part
+   * of the exported API for library1, but by default API Extractor would generate a .d.ts rollup that explicitly
+   * imports library2.  To avoid this, we can specify:
+   *
+   *   "bundledPackages": [ "library2" ],
+   *
+   * This would direct API Extractor to embed those types directly in the .d.ts rollup, as if they had been
+   * local files for library1.
+   */
+  "bundledPackages": [ "@textile/buckets", "@textile/threads-id", "@textile/threads-client" ],
```
andrewxhill, line 63: here's where we define the other libs we want to tow along

docs/README.md
```diff
@@ -0,0 +1,197 @@
+# @textile/hub
```
andrewxhill, line 1: this is auto copied from the hub readme.

docs/_sidebar.md
```diff
@@ -0,0 +1,16 @@
+- Getting Started
```
andrewxhill, line 1: this is manual unfortunately

package.json
```diff
@@ -3,18 +3,22 @@
...
   "scripts": {
+    "docs:markdown": "api-documenter markdown --input-folder=temp",
+    "docs:html": "npm run docs:markdown && cp -r markdown/* docs/ && cp packages/hub/README.md ./docs/",
+    "build:docs": "npm run docs:markdown && npm run docs:html",
+    "build:extractor": "api-extractor run --local --verbose",
+    "build:lerna": "lerna run build",
     "bootstrap": "lerna bootstrap",
     "compile": "lerna run compile",
-    "build": "lerna run build",
+    "build": "npm run build:lerna && npm run build:extractor && npm run build:docs",
```
andrewxhill, line 13: okay, so docs will regenerate anytime it's built now. 

the docs folder is checked into the repo. so we should be able to just change our docs deployment choice and it'll be live.

packages/buckets/src/buckets.ts
```diff
@@ -21,6 +21,20 @@ export interface PushPathResult {
 
 /**
  * Buckets is a web-gRPC wrapper client for communicating with the web-gRPC enabled Textile Buckets API.
+ * @example
```
andrewxhill, line 24: if we add some of these inline examples to the client lib, they should start getting picked up too

packages/hub/src/users.ts
```diff
@@ -77,4 +70,7 @@ Client.prototype.listThreads = async function (ctx?: Context) {
   })
 }
 
-export { Client }
+/**
+ * Clients is a web-gRPC wrapper client for communicating with the web-gRPC enabled Threads API.
+ */
+export { Client, UserAuth, KeyInfo }
```
andrewxhill, line 76: i want to ensure that these types (and maybe others) are available to hub users. i'll likely split them into a new js-textile package so i can document them easier here too

package.json
carsonfarmer: hmmm, this is a bummer (having the docs folder checked into the repo), is this just for the 'static' assets? Not the generated docs right?

packages/buckets/src/buckets.ts
carsonfarmer: nice

Issue #270, opened 2020-06-26
Add method for `Bucket.fromClient` to reuse the context created in a db client


andrewxhill:
Right now, in docs and elsewhere we do something like.

```typescript
// Setup the client with all the token and session data, as variable `client`
// Now, start a bucket client
const buckets = new Buckets(client.context)
```

We've moved away from context in all other places. So it feels like a better api here would just be,

```typescript
const buckets = Buckets.fromClient(client)
```

PR #271, opened 2020-06-26
feat(init): adds method to init from client directly

andrewxhill:
fixes https://github.com/textileio/js-hub/issues/270

### DIFF
--- a/packages/buckets/package.json
+++ b/packages/buckets/package.json
@@ -32,6 +32,7 @@
     "@textile/buckets-grpc": "^1.0.8",
     "@textile/context": "^0.4.7",
     "@textile/security": "^0.1.4",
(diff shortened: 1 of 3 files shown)


packages/buckets/package.json
```diff
@@ -32,6 +32,7 @@
     "@textile/buckets-grpc": "^1.0.8",
     "@textile/context": "^0.4.7",
     "@textile/security": "^0.1.4",
+    "@textile/threads-client": "^0.6.13",
```
andrewxhill, line 35: i think this used to be an issue because it would pull along a bunch of crypto stuff. i think it's fine now

carsonfarmer: Also, since you're only using the types here, it _shouldn't_ bring in anything to bundles etc.

Issue #275, opened 2020-06-28
Tighten up bucket starting flow


andrewxhill:
Here's an example of a normal flow to create/re-enter a bucket in js

```typescript
getBucketThreadID = async () => {
  try {
    var storedThread = localStorage.getItem("bucket-thread-id")
    if (storedThread === null) {
      throw new Error('No thread id')
    }
    const = ThreadID.fromString(storedThread)
    return restored
  }
  catch (e) {
    const newThread = ThreadID.fromRandom()
    localStorage.setItem("bucket-thread-id", newThread.toString())
    return threadID
  }
}

const client = await Client.withUserAuth(this.userAuth)
await client.getToken(this.identity)
const threadID = await getBucketThreadID()
client.open(threadID, "home-for-buckets")
const buck = Buckets.fromClient(client);
this.buckets = buck.init("memes");
```

It would be really nice if we could just do something like, where the desired target thread is just passed as a name.

```typescript
const buck = await Buckets.withUserAuth(this.userAuth, "home-for-buckets")
this.buckets = buck.init("memes");
```

PR #278, opened 2020-06-29
feat(buckets): adds simplified bucket open method without thread requirements

andrewxhill:
fixes https://github.com/textileio/js-hub/issues/275


packages/buckets/src/buckets.spec.ts
```diff
@@ -26,15 +26,21 @@ describe('Buckets...', () => {
 
   before(async () => {
     const user = await signUp(ctx, addrGatewayUrl, sessionSecret)
-    const id = ThreadID.fromRandom()
-    const db = new Client(ctx.withSession(user.user?.session).withThreadName('buckets'))
-    await db.newDB(id)
```
andrewxhill, line 31: removing from tests... which means user don't have to do this anymore either

packages/buckets/src/buckets.spec.ts
```diff
@@ -26,15 +26,21 @@ describe('Buckets...', () => {
...
+    ctx.withSession(user.user?.session)
   })
 
-  it('should init a new bucket', async () => {
+  it('should open a bucket by name without thread info', async () => {
+    const root = await client.open('initbuck')
```
andrewxhill, line 33: open method can just take the name of the bucket. it will try to use the default 'buckets' thread if no thread name or thread id is provided. otherwise it will create it

packages/buckets/src/buckets.spec.ts
```diff
@@ -26,15 +26,21 @@ describe('Buckets...', () => {
...
+    expect(root).to.have.ownProperty('updatedat')
+  })
+
+  it('should init a new bucket on open thread', async () => {
     // Check that we're empty
     const list = await client.list()
-    expect(list).to.have.length(0)
+    expect(list).to.have.length(1)
     // Now initialize a bucket
     buck = await client.init('mybuck')
```
andrewxhill, line 45: init could just go away, but it still works fine the same way as before too. it's just complicated

packages/buckets/src/buckets.ts
```diff
@@ -6,7 +6,8 @@ import { EventIterator } from 'event-iterator'
 import { grpc } from '@improbable-eng/grpc-web'
 import { ContextInterface, Context, defaultHost } from '@textile/context'
 import { UserAuth, KeyInfo } from '@textile/security'
-import { Client } from '@textile/threads-client'
+import { ThreadID } from '@textile/threads-id'
+import { Client } from '@textile/hub-threads-client'
```
andrewxhill, line 10: i needed to move this definition so as to ensure no circular dependencies. but now i'm realizing i could probably just import from threads-client instead of the hub threads. it could still be a good move just to isolate this stuff anyway, let me know what you think

packages/buckets/src/buckets.ts
```diff
@@ -90,6 +97,47 @@ export class Buckets {
     return new Buckets(client.context)
   }
 
+  /**
+   * Open a new / existing bucket by bucket name and ThreadID
+   * @param name name of bucket
+   * @param threadID id of thread where bucket is stored
+   */
+  async open(name: string, threadName?: string, threadID?: ThreadID) {
+    if (threadID) {
```
andrewxhill, line 106: allows for directly picking threadid. but i'm going to move examples to just using default 'buckets' most of the time. so just `buckets.open(bucket_name)`

packages/buckets/src/buckets.ts
carsonfarmer: Assuming in buckets we don't need to augmented APIs that come with the User module, then yes, it'd be better to just import from threads-client directly. I'd like to keep things mostly contained, plus the more libs on npm there are, the more chances for someone to pull in the wrong thing.

andrewxhill: So I do use the augmented API, see line 125. So I assume I do need it here.

carsonfarmer:
This seems reasonable @andrewxhill. I'd rather _not_ have another sub-package that only just exports threads-client again. Is the circular dep only caused by tests? I don't see it off the top of my head.

Note also that the test failure appears to be due to an overly specified test (my doing). Perhaps while you're in there you can just update to reflect the fact that the error just includes "nope.jpg" or something?

andrewxhill:
The new client is required, see comment above. Re. more packages, I don't think that's a real concern anymore given the total number of packages. it just comes down to documentation, that's why the other week i was trimming everything down to two possible packages `@textile/hub` and `@textile/threads`. database is still a dangling library, but will move it once i have a good example that works and i can verify in the process.

PR #292, opened 2020-07-02
chore(deps): pulls in updated auth helper; fixes crypto

andrewxhill:


packages/buckets/src/spec.util.ts
```diff
@@ -47,11 +49,13 @@ export const signUp = (ctx: ContextInterface, addrGatewayUrl: string, sessionSec
       req.setEmail(email)
       req.setUsername(username)
       const client = new APIClient(ctx.host, { transport: ctx.transport, debug: ctx.debug })
-      client.signup(req, ctx.toMetadata(), (err: ServiceError | null, message: pb.SignupReply | null) => {
-        if (err) reject(err)
-        resolve({ user: message?.toObject(), username, email })
+      ctx.toMetadata().then((meta) => {
```
andrewxhill, line 52: this returns promises now. so the bulk of this PR was just updating them to handle that

PR #294, opened 2020-07-02
chore(deps): pulls in updated auth helper; fixes crypto

andrewxhill:
Reverts textileio/js-hub#293


scripts/change-client-imports-to-hub
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
andrewxhill, line 1: worth pointing this out @carsonfarmer. it gets run during the docs build to ensure examples show user how to import client from textile/hub not textile/threads. this also seems to sort out that type error i was seeing...

PR #320, opened 2020-07-07
chore(nits): clean up use of client internally

andrewxhill:


packages/buckets/src/buckets.ts
```diff
@@ -52,6 +52,7 @@ export interface PushPathResult {
  * ```
  */
 export class Buckets {
+  private _client: Client
```
andrewxhill, line 55: this is starting to mirror more closely buckd, where buck has most of threads capabilities. wanted to simplify the context moving for buckets, so just hanging onto the private instance here.

PR #332, opened 2020-07-12
fix(api): bucket methods without crypto

andrewxhill:


config/api-extractor.json
```diff
@@ -60,7 +60,7 @@
    * This would direct API Extractor to embed those types directly in the .d.ts rollup, as if they had been
    * local files for library1.
    */
-  "bundledPackages": [ "@textile/buckets", "@textile/hub-threads-client", "@textile/threads-id", "@textile/threads-client", "@textile/security" ],
+  "bundledPackages": [ "@textile/buckets", "@textile/hub-threads-client", "@textile/threads-id", "@textile/threads-client", "@textile/security", "@textile/core" ],
```
andrewxhill, line 63: note to remove again

packages/buckets/src/index.ts
```diff
@@ -1,2 +1,23 @@
...
+  ArchiveStatusReply,
+  ArchiveWatchReply,
+  InitReply,
+  ListPathItem,
+  ListPathReply,
+  ListReply,
+  PullIpfsPathReply,
+  PullPathReply,
+  PushPathReply,
+  SetPathReply,
```
andrewxhill, line 22: got explicit in my response typing

packages/buckets/src/api/index.ts
```diff
@@ -0,0 +1,523 @@
+import log from 'loglevel'
```
andrewxhill, line 1: i'm expecting that i could `import { specifics } from '@textile/buckets/api'`, so this has no crypto. however... that always confuses me a bit. so maybe that dep would show up unless i make this a standalone package...

andrewxhill: let me know your thoughts here @carsonfarmer

packages/buckets/src/buckets.ts
```diff
@@ -225,37 +240,28 @@ export class Buckets {
...
    */
-  async listPath(key: string, path: string, ctx?: ContextInterface) {
+  async listPath(key: string, path: string): Promise<ListPathReply.AsObject> {
     logger.debug('list path request')
-    const req = new pb.ListPathRequest()
-    req.setKey(key)
-    req.setPath(path)
-    const res: pb.ListPathReply = await this.unary(API.ListPath, req, ctx)
-    return res.toObject()
+    return bucketsListPath(this, key, path)
```
andrewxhill, line 255: here's an example using the function method but adding it here to the stateful class so that documentation will be clear

PR #376, opened 2020-07-28
feat(dirs): adds full dir path tree and list

andrewxhill:
will walk the full path tree and return a complete object or list of paths

e.g. 

```
     [
       'mybuck',
       'mybuck/.textileseed',
       'mybuck/dir1',
       'mybuck/dir1/file1.jpg',
       'mybuck/path',
       'mybuck/path/to',
       'mybuck/path/to/file2.jpg'
     ]
```

### DIFF
--- a/etc/hub.api.md
+++ b/etc/hub.api.md
@@ -87,8 +87,7 @@ export class Buckets extends BucketsGrpcClient {
     links(key: string): Promise<LinksReply.AsObject>;
     list(): Promise<Root.AsObject[]>;
     listIpfsPath(path: string): Promise<ListPathItem.AsObject | undefined>;
-    listPath(key: string, path: string): Promise<ListPathReply.AsObject>;
-    listPathRecursive(key: string, path: string): Promise<ListPathRecursive>;
+    listPath(key: string, path: string, recursive?: boolean): Promise<ListPathReply.AsObject>;
     listPathRecursiveFlat(key: string, path: string, dirs?: boolean): Promise<ListPathRecursiveFlat>;
     open(name: string, threadName?: string, isPrivate?: boolean, threadID?: string): Promise<Root.AsObject | undefined>;
     pullIpfsPath(path: string, opts?: {
@@ -174,6 +173,9 @@ export function bucketsRemovePath(api: BucketsGrpcClient, key: string, path: str
 
 // @public
 export function bucketsRoot(api: BucketsGrpcClient, key: string, ctx?: ContextInterface): Promise<Root.AsObject | undefined>;
+
+// @public
+export function bufToArray(chunk: Buffer, size?: number): Buffer[];
 
 // @public
 export class Client {
@@ -339,10 +341,16 @@ export { ListIpfsPathReply }
 export { ListPathItem }
 
 // @public (undocumented)
-export type ListPathRecursive = ReturnType<typeof utilListPathRecursive>;
+export type ListPathRecursive = ReturnType<typeof listPathRecursive>;
+
+// @public
+export function listPathRecursive(grpc: BucketsGrpcClient, bucketKey: string, path: string): Promise<ListPathReply.AsObject>;
 
 // @public (undocumented)
-export type ListPathRecursiveFlat = ReturnType<typeof utilListPathRecursiveFlat>;
+export type ListPathRecursiveFlat = ReturnType<typeof listPathRecursiveFlat>;
+
+// @public
+export function listPathRecursiveFlat(grpc: BucketsGrpcClient, bucketKey: string, path: string, dirs?: boolean): Promise<string[]>;
 
 export { ListPathReply }
 
@@ -441,15 +449,6 @@ export type UserAuth = {
     msg: string;
     token?: string;
 };
-
-// @public
-export function utilBufToArray(chunk: Buffer, size?: number): Buffer[];
-
-// @public
-export function utilListPathRecursive(grpc: BucketsGrpcClient, bucketKey: string, path: string): Promise<ListPathItem.AsObject | undefined>;
-
-// @public
-export function utilListPathRecursiveFlat(grpc: BucketsGrpcClient, bucketKey: string, path: string, dirs?: boolean): Promise<string[]>;
 
 // @public
 export enum Variant {
--- a/etc/hub.api.md
+++ b/etc/hub.api.md
@@ -81,15 +81,23 @@ export class Buckets extends BucketsGrpcClient {
         id: string | undefined;
         msg: string;
     }, err?: Error) => void): Promise<() => void>;
+    getOrInitBucket(name: string, threadName?: string, isPrivate?: boolean, threadID?: string): Promise<{
+        root?: Root.AsObject;
+        threadID?: string;
+    }>;
     getToken(identity: Identity): Promise<string>;
     getTokenChallenge(publicKey: string, callback: (challenge: Uint8Array) => Uint8Array | Promise<Uint8Array>): Promise<string>;
     init(name: string, isPrivate?: boolean): Promise<InitReply.AsObject>;
     links(key: string): Promise<LinksReply.AsObject>;
     list(): Promise<Root.AsObject[]>;
     listIpfsPath(path: string): Promise<ListPathItem.AsObject | undefined>;
-    listPath(key: string, path: string, recursive?: boolean): Promise<ListPathReply.AsObject>;
-    listPathRecursiveFlat(key: string, path: string, dirs?: boolean): Promise<ListPathRecursiveFlat>;
-    open(name: string, threadName?: string, isPrivate?: boolean, threadID?: string): Promise<Root.AsObject | undefined>;
+    listPath(key: string, path: string, depth?: number): Promise<ListPathReply.AsObject>;
+    listPathRecursiveFlat(key: string, path: string, dirs?: boolean, depth?: number): Promise<ListPathRecursiveFlat>;
+    // @deprecated
+    open(name: string, threadName?: string, isPrivate?: boolean, threadID?: string): Promise<{
+        root?: Root.AsObject;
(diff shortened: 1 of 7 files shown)


packages/buckets/src/api/index.ts
```diff
@@ -239,6 +239,7 @@ export async function bucketsListIpfsPath(
  * @param opts Options to control response stream. Currently only supports a progress function.
  * @remarks
  * This will return the resolved path and the bucket's new root path.
+ * Data must be broken into <4mb chunks. See bufToArray() for help.
```
carsonfarmer, line 242: 👍

packages/buckets/src/buckets.ts
```diff
@@ -253,6 +254,46 @@ export class Buckets extends BucketsGrpcClient {
...
+   * //   'mybuck/.textileseed',
+   * //   'mybuck/dir1',
+   * //   'mybuck/dir1/file1.jpg',
+   * //   'mybuck/path',
+   * //   'mybuck/path/to',
+   * //   'mybuck/path/to/file2.jpg'
+   * // ]
+   * ```
+   */
+  async listPathRecursiveFlat(key: string, path: string, dirs = true): Promise<ListPathRecursiveFlat> {
```
carsonfarmer, line 292: This is nice! But why not just `listPath(key: string, path: string, opts?: { dirs: boolean, recursive: boolean, flatten: boolean })`

packages/buckets/src/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  const result = []
+  const len = chunk.length
+  let i = 0
+  while (i < len) {
+    result.push(chunk.slice(i, (i += size)))
+  }
+  return result
+}
+
+export type ListPathRecursive = ReturnType<typeof utilListPathRecursive>
```
carsonfarmer, line 19: Very nice!

packages/buckets/src/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+  return result
+}
+
+export type ListPathRecursiveFlat = ReturnType<typeof utilListPathRecursiveFlat>
+
+/**
+ * utilPathsList returns a string array of all paths in a bucket
+ */
+export async function utilListPathRecursiveFlat(grpc: BucketsGrpcClient, bucketKey: string, path: string, dirs = true) {
```
carsonfarmer, line 60: Why the util prefix? Is this a convention we should be using?

packages/buckets/src/buckets.ts
andrewxhill: i just wasn't sold on the various return types.  i can definitely get two of them into one, normal and recursive. but the flat method is a bit different, what do you think?

andrewxhill: hmm... i also don't want to make the `dirs` option an option on the `listPath` method. so there is a bit of a mismatch there ("i used list path with, opts: {dirs: false}, but i'm still getting dirs?" type questions pretty quick)

carsonfarmer: I'm sold

packages/buckets/src/utils.ts
andrewxhill: oh, rm btw

packages/buckets/src/buckets.ts
```diff
@@ -103,22 +138,22 @@ export class Buckets extends BucketsGrpcClient {
...
    *     const buckets = Buckets.withUserAuth(auth)
-   *     await buckets.open(name)
-   *     return buckets
+   *     const { root, threadID } = await buckets.getOrInitBucket(name)
+   *     return { buckets, root, threadID }
    * }
    * ```
    */
-  async open(
+  async getOrInitBucket(
```
andrewxhill, line 146: snuck this in so that if we add the api into go buckets, i can just replace this internal logic with the endpoint. deprecates `open`

packages/buckets/src/buckets.ts
```diff
@@ -135,17 +170,18 @@ export class Buckets extends BucketsGrpcClient {
...
+        threadID = newId.toString()
+        this.withThread(threadID)
       }
     }
 
     const roots = await this.list()
     const existing = roots.find((bucket) => bucket.name === name)
     if (existing) {
-      return existing
+      return { root: existing, threadID }
```
andrewxhill, line 181: as per convo about what's needed in that method, started returning the threadid

packages/buckets/src/buckets.ts
```diff
@@ -247,10 +283,42 @@ export class Buckets extends BucketsGrpcClient {
    * Returns information about a bucket path.
    * @param key Unique (IPNS compatible) identifier key for a bucket.
    * @param path A file/object (sub)-path within a bucket.
+   * @param depth (optional) will walk the entire bucket to target depth (default = 1)
    */
-  async listPath(key: string, path: string): Promise<ListPathReply.AsObject> {
+  async listPath(key: string, path: string, depth = 1): Promise<ListPathReply.AsObject> {
```
andrewxhill, line 288: switched it to depth, so that it's not going to walk the full directory all the time. instead you need to manually tell it how deep to go. feels a bit safer than unbounded recursive, thoughts?

packages/buckets/src/buckets.ts
```diff
@@ -247,10 +283,42 @@ export class Buckets extends BucketsGrpcClient {
...
+   * //   'mybuck/.textileseed',
+   * //   'mybuck/dir1',
+   * //   'mybuck/dir1/file1.jpg',
+   * //   'mybuck/path',
+   * //   'mybuck/path/to',
+   * //   'mybuck/path/to/file2.jpg'
+   * // ]
+   * ```
+   */
+  async listPathRecursiveFlat(key: string, path: string, dirs = true, depth = 5): Promise<ListPathRecursiveFlat> {
```
andrewxhill, line 319: same here but higher default

etc/hub.api.md
```diff
@@ -81,14 +81,23 @@ export class Buckets extends BucketsGrpcClient {
...
     init(name: string, isPrivate?: boolean): Promise<InitReply.AsObject>;
     links(key: string): Promise<LinksReply.AsObject>;
     list(): Promise<Root.AsObject[]>;
     listIpfsPath(path: string): Promise<ListPathItem.AsObject | undefined>;
-    listPath(key: string, path: string): Promise<ListPathReply.AsObject>;
-    open(name: string, threadName?: string, isPrivate?: boolean, threadID?: string): Promise<Root.AsObject | undefined>;
+    listPath(key: string, path: string, depth?: number): Promise<ListPathReply.AsObject>;
+    listPathFlat(key: string, path: string, dirs?: boolean, depth?: number): Promise<ListPathFlat>;
+    // @deprecated
+    open(name: string, threadName?: string, isPrivate?: boolean, threadID?: string): Promise<{
```
carsonfarmer, line 97: 👍

packages/buckets/src/buckets.ts
```diff
@@ -26,26 +26,42 @@ import {
...
+ *   return { buckets, bucketKey }
+ * }
+ * ```
+ *
+ * @example
+ * Print the links for the bucket
+ * ```typescript
+ * import { Buckets } from '@textile/hub'
+ *
+ * // This method requires that you run "open" or have specified "withThread"
```
carsonfarmer, line 61: nit: doc update because we're deprecating "open".

packages/buckets/src/buckets.ts
```diff
@@ -54,6 +70,8 @@ const logger = log.getLogger('buckets')
  * ```typescript
  * import { Buckets } from '@textile/hub'
  *
+ * // This method requires that you already specify the Thread containing
+ * // the bucket with buckets.withThread(<thread name>).
```
carsonfarmer, line 74: Wondering if we should model this more like the Go lib, and actually include the with* in the constructor?

```
const buckets = new Buckets(withThread(...))
```

Maybe more difficult here because we might not already know the Thread, but this would be optional anyway.

packages/buckets/src/buckets.ts
carsonfarmer: Is setting depth to 0 going to be essential infinite? This seems reasonable to me.

packages/buckets/src/utils.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
+import { Buffer } from 'buffer'
+import { ListPathItem, ListPathReply } from '@textile/buckets-grpc/buckets_pb'
+import { bucketsListPath, BucketsGrpcClient } from './api'
+/**
+ * utilBufToArray converts a buffer into <4mb chunks for use with grpc API
+ * @param chunk an input Buffer
+ */
+export function bufToArray(chunk: Buffer, size = 1024 * 1024 * 3) {
```
carsonfarmer, line 9: Since Buffers _are_ Uint8Arrays, it might be more generally useful to have this as Uint8Array. There's nothing in the code that would preclude this. You could still _return_ an array of buffers if you wanted?

packages/buckets/src/buckets.ts
andrewxhill: We never call 'new Buckets(...)` for anything though. you almost always call, `Buckets.withUserAuth(...)` which means you must set the threadId after in almost all cases. and even more obvious is to just call the `getOrInit` which never makes you deal with the thread. i just wanted to include the note where you are commenting in case people go down this adventurous path

packages/buckets/src/buckets.ts
andrewxhill: not really. depth 0 will return the same as depth 1 because the api always returns leafs within 1 order down. -1 is unbounded.

packages/buckets/src/utils.ts
andrewxhill: updated

PR #382, opened 2020-07-30
fix(api): improvements based on using in ci build; docs adds

andrewxhill:


etc/hub.api.md
```diff
@@ -72,7 +72,7 @@ export class Buckets extends BucketsGrpcClient {
     list(): Promise<Root.AsObject[]>;
     listIpfsPath(path: string): Promise<ListPathItem.AsObject | undefined>;
     listPath(key: string, path: string, depth?: number): Promise<ListPathReply.AsObject>;
-    listPathFlat(key: string, path: string, dirs?: boolean, depth?: number): Promise<ListPathFlat>;
+    listPathFlat(key: string, path: string, dirs?: boolean, depth?: number): Promise<Array<string>>;
```
andrewxhill, line 75: didn't need a special type obv

packages/buckets/src/buckets.spec.ts
```diff
@@ -145,16 +145,17 @@ describe('Buckets...', () => {
...
-    //   'mybuck/path',
-    //   'mybuck/path/to',
-    //   'mybuck/path/to/file2.jpg'
+    //   '.textileseed',
+    //   'dir1',
+    //   'dir1/file1.jpg',
+    //   'path',
+    //   'path/to',
+    //   'path/to/file2.jpg'
     // ]
```
andrewxhill, line 154: having the flat paths prefixed by the buck name wasn't helpful when building with this list

packages/buckets/src/buckets.ts
```diff
@@ -77,6 +77,31 @@ const logger = log.getLogger('buckets')
  *     return roots.find((bucket) => bucket.name === bucketName)
  * }
  * ```
+ *
+ * @example
+ * Push an folder in node.js
```
andrewxhill, line 82: adds streaming file pushes for folder sync via nodejs

packages/buckets/src/buckets.ts
```diff
@@ -348,6 +373,31 @@ export class Buckets extends BucketsGrpcClient {
    *    return await buckets.pushPath(bucketKey!, 'index.html', file)
    * }
    * ```
+   * 
+   * @example
+   * Push an folder in node.js
```
andrewxhill, line 378: duplicate example. with example checking though, shouldn't be issue keeping updated

packages/buckets/src/utils.ts
```diff
@@ -24,10 +23,10 @@ export async function listPathRecursive(
   path: string,
   depth: number,
   currentDepth = 0,
-) {
+): Promise<ListPathReply.AsObject> {
```
andrewxhill, line 26: better build

packages/buckets/src/utils.ts
```diff
@@ -41,33 +40,36 @@ export async function listPathRecursive(
...
-  // Only push if dirs included or not a dir
-  if (dirs || !tree.isdir) result.push(fp)
-  if (tree.isdir) {
-    for (const item of tree.itemsList) {
-      const downtree = await treeToPaths(item, fp, dirs, depth)
+  for (const item of tree) {
+    const newPath = path === '' ? `${item.name}` : `${path}/${item.name}`
+    if (dirs || !item.isdir) result.push(newPath)
+    if (item.isdir && (currentDepth < depth || depth === -1)) {
+      const downtree = await treeToPaths(item.itemsList, newPath, dirs, depth, currentDepth + 1)
```
andrewxhill, line 55: just changes since the buck name isn't prefixing the paths now

website/sidebars.json
```diff
@@ -23,14 +23,15 @@
       "hub.buckets.withuserauth",
       "hub.buckets.withkeyinfo",
       "hub.buckets.gettoken",
-      "hub.buckets.getOrInit",
+      "hub.buckets.getorinit",
```
andrewxhill, line 26: ooop, lowwercase only

packages/buckets/src/utils.ts
carsonfarmer: Yeh, we should just _always_ do this for external APIs. I have a linter setting coming in that'll make sure we do.

PR #386, opened 2020-07-31
feat(mailbox): adding user api

andrewxhill:


packages/users/src/users.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  async getThread(name: string): Promise<GetThreadReply.AsObject> {
+    return getThread(this, name)
+  }
+
+  async setupMailbox(): Promise<{ mailboxID: Uint8Array | string }> {
+    return setupMailbox(this)
+  }
+
+  async sendMessage(
```
andrewxhill, line 93: This is intended to match the API found here, https://github.com/textileio/textile/blob/5d8532b244aa0628ce401c11cf42c8e0953d6aeb/api/users/client/client.go#L128 but the libp2p crypto types don't have all the necessary functions to fulfill this.

packages/users/src/api/index.ts
```diff
@@ -0,0 +1,323 @@
...
+  const req = new DeleteMessageRequest()
+  req.setId(id)
+  await api.unary(API.DeleteSentboxMessage, req, ctx)
+  return
+}
+
+/**
+ * @internal
+ */
+export function watchMailbox(
```
andrewxhill, line 287: new listen method for mailbox @carsonfarmer

packages/users/src/users.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+      rec = await user1.listSentboxMessages()
+      expect(rec.length).to.equal(0)
+    })
+    it('mailboxID should exist', async () => {
+      const user1 = new Users(user1Ctx)
+      const mailboxID = await user1.getMailboxID()
+      expect(mailboxID).to.not.be.undefined
+    })
+
+    it('inbox listen should return new messages', (done) => {
```
andrewxhill, line 313: new listen tests, this one 'til the end @carsonfarmer

etc/hub.api.md
```diff
@@ -48,10 +64,12 @@ export { ArchiveStatusReply }
 
 export { ArchiveWatchReply }
 
+// Warning: (ae-incompatible-release-tags) The symbol "Buckets" is marked as @public, but its signature references "GrpcAuthentication" which is marked as @internal
```
carsonfarmer, line 67: Is this an issue?

packages/buckets/src/api/index.ts
```diff
@@ -55,51 +55,84 @@ export interface PushPathResult {
...
-   * @param context The context to use for interacting with the APIs. Can be modified later.
-   */
-  constructor(public context: ContextInterface = new Context(), debug = false) {
-    this.serviceHost = context.host
-    this.rpcOptions = {
-      transport: WebsocketTransport(),
-      debug,
-    }
+/**
+ * Response from bucket links query.
```
carsonfarmer, line 59: I know is a pain to maintain these outside the gRPC types, but wow does this improve the DX considerably. Too bad there isn't a way to derive these automatically from the gRPC responses?

packages/buckets/src/buckets.ts
```diff
@@ -103,34 +99,73 @@ const logger = log.getLogger('buckets')
  * }
  * ```
  */
-export class Buckets extends BucketsGrpcClient {
+export class Buckets extends GrpcAuthentication {
   /**
-   * Creates a new gRPC client instance for accessing the Textile Buckets API.
-   * @param auth The user auth object.
+   * {@inheritDoc @textile/hub#GrpcAuthentication.copyAuth}
```
carsonfarmer, line 104: Cool!

packages/crypto/src/keypair.ts
```diff
@@ -0,0 +1,298 @@
...
+  return new TextDecoder().decode(encoded)
+}
+
+/**
+ * Default implementation of the {@link Public} interface, with encryption
+ * support.
+ * In theory, RSA, ed25519, and secp256k1 key types (and more) should be
+ * supported, although currently only ed25519 has full verify and encrypt
+ * capabilities.
+ * @privateRemarks Support additional key types by default. For now, sticking to ed25519
```
carsonfarmer, line 23: Cool!

packages/crypto/src/keypair.ts
```diff
@@ -0,0 +1,298 @@
...
+ * import { PrivateKey } from '@textile/hub'
+ *
+ * function example() {
+ *   const identity = PrivateKey.fromRandom()
+ *   const publicKey = identity.public.toString()
+ *   return publicKey
+ * }
+ * ```
+ *
+ * @example
```
carsonfarmer, line 41: Love these!

packages/users/src/api/index.ts
carsonfarmer: Sweet, yup looks good. One day we'll get those Buffer.from's out of everything.

etc/hub.api.md
andrewxhill: I hid it intentionally so it doesn't make the docs unclear. This warning just means that where docs reference it now it wont be linked to anything. So in particular, when it says Users class is an instance of this, it's just plaintext

PR #396, opened 2020-08-04
Additional docs strings + details

carsonfarmer:
Some doc strings for your PR @andrewxhill

### DIFF
--- a/packages/crypto/package.json
+++ b/packages/crypto/package.json
@@ -31,6 +31,7 @@
   "author": "Textile",
   "license": "MIT",
   "devDependencies": {
+    "libp2p-crypto": "^0.17.8",
     "rimraf": "^3.0.2",
     "typescript": "^3.9.7"
   },
--- a/packages/crypto/src/identity.ts
+++ b/packages/crypto/src/identity.ts
@@ -43,12 +43,6 @@ export interface Public {
   verify(data: Uint8Array, sig: Uint8Array): Promise<boolean>
 
   /**
-   * Return the base32-encoded multibase string of the `bytes` of this public
-   * key. Useful for encoding the key for sharing etc.
-   */
-  toString(): string
-
-  /**
    * Return the protobuf-encoded bytes of this public key.
    */
   bytes: Uint8Array
--- a/packages/crypto/src/identity.ts
+++ b/packages/crypto/src/identity.ts
@@ -1,18 +1,55 @@
-export type Private = Identity
-
 /**
  * Identity represents an entity capable of signing a message.
- * This is a simple 'private key' interface that must be capable of returning the associated public key for
- * verification. In many cases, this will just be a private key, but callers can use any setup that suits their needs.
- * The interface is currently modeled after @textile/threads-crypto PrivateKeys.
+ * This is a simple "private key" interface that must be capable of returning
+ * the associated "public key" for verification. In many cases, the Identity
+ * will just be a private key, but callers can use any setup that suits their
+ * needs. The interface is currently modeled after `libp2p-crypto` keys.
+ * It is meant to be implemented by external libraries, though in practice,
+ * implementations are provided by the `@textile/crypto` library.
+ * @alias Private
  */
 export interface Identity {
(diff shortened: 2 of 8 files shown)


packages/crypto/package.json
```diff
@@ -31,6 +31,7 @@
   "author": "Textile",
   "license": "MIT",
   "devDependencies": {
+    "libp2p-crypto": "^0.17.8",
```
carsonfarmer, line 34: adding this here so we can check that they also satisfy our required interface

packages/crypto/src/identity.ts
```diff
@@ -1,18 +1,49 @@
-export type Private = Identity
-
 /**
  * Identity represents an entity capable of signing a message.
- * This is a simple 'private key' interface that must be capable of returning the associated public key for
- * verification. In many cases, this will just be a private key, but callers can use any setup that suits their needs.
- * The interface is currently modeled after @textile/threads-crypto PrivateKeys.
+ * This is a simple "private key" interface that must be capable of returning
```
carsonfarmer, line 3: better docs!

packages/crypto/src/identity.ts
```diff
@@ -1,18 +1,49 @@
...
+ */
 export interface Public {
+  /**
+   * Verifies the signature for the data and returns true if verification
+   * succeeded or false if it failed.
+   * @param data The data to use for verification.
+   * @param sig The signature to verify.
+   */
   verify(data: Uint8Array, sig: Uint8Array): Promise<boolean>
-  toString(): string
```
carsonfarmer, line 16: note that I'm removing this. because this is an implementation detail, and is specific to our usage of the keys. they'll all have a toString, but we can't rely on it for marshaling, so we'll leave it out here

packages/crypto/src/keypair.spec.ts
```diff
@@ -32,22 +51,32 @@ describe('Keypair', () => {
...
-
       // I encrypt it...
       const ciphertext = await encrypt(msg, pubKey.pubKey) // Don't use bytes!
-
       // They decrypt it...
       const plaintext = await decrypt(ciphertext, privKey.privKey)
+      expect(plaintext).to.deep.equal(msg)
+    })
 
+    it('should be able to encrypt/decrypt using classes', async () => {
```
carsonfarmer, line 67: added this just as an alternative flow and a good example of using the classes to marshal/unmarshal

packages/crypto/src/keypair.ts
```diff
@@ -2,57 +2,112 @@ import nacl from 'tweetnacl'
...
+/**
+ * Encode the given PrivateKey to its base-32 encoded multibase representation.
+ * @param key The input PrivateKey.
+ */
 export function privateKeyToString(key: PrivateKey) {
   const encoded = multibase.encode('base32', key.bytes as Buffer)
   return new TextDecoder().decode(encoded)
 }
 
+/**
```
carsonfarmer, line 17: better docs

packages/crypto/src/keypair.ts
```diff
@@ -2,57 +2,112 @@ import nacl from 'tweetnacl'
...
+ * Default implementation of the {@link Public} interface, with encryption
+ * support.
+ * In theory, RSA, ed25519, and secp256k1 key types (and more) should be
+ * supported, although currently only ed25519 has full verify and encrypt
+ * capabilities.
+ * @todo Support additional key types by default. For now, sticking to ed25519
+ * keeps our bundle smaller, but if needed, additional types can be added via
+ * libp2p-crypto or tweetnacl.
+ * @todo Separate out the generic PublicKey interface to include signing and
```
andrewxhill, line 26: let's try to stick to the api-extractor supported tags, https://api-extractor.com/pages/tsdoc/doc_comment_syntax/ to ensure that rendering works as expected

packages/crypto/src/keypair.spec.ts
andrewxhill: updating on my branch though

PR #458, opened 2020-08-25
feat(api): automatically chunk input

andrewxhill:


packages/buckets/src/api/index.ts
```diff
@@ -403,10 +405,13 @@ export async function bucketsPushPath(
...
-          part.setChunk(chunk as Buffer)
-          client.send(part)
+        const process = await block({ size: 4096, noPad: true })
+        for await (const chunk of process(source.content)) {
+          for (const buf of chunk._bufs) {
+            const part = new PushPathRequest()
+            part.setChunk(buf as Buffer)
+            client.send(part)
+          }
         }
```
andrewxhill, line 415: chunk change

packages/buckets/src/api/index.ts
```diff
@@ -403,10 +405,13 @@ export async function bucketsPushPath(
...
-          const part = new PushPathRequest()
-          part.setChunk(chunk as Buffer)
-          client.send(part)
+        const process = await block({ size: 4096, noPad: true })
+        for await (const chunk of process(source.content)) {
+          for (const buf of chunk._bufs) {
+            const part = new PushPathRequest()
+            part.setChunk(buf as Buffer)
+            client.send(part)
+          }
```
carsonfarmer, line 414: This makes me slightly nervous, as its not really how we should consume the bl object. Since its already doing the chucking for us, what about (https://www.npmjs.com/package/bl#blslice-start--end--):

```suggestion
          // New Buffer object containing bytes of chunk, even if split over multiple sub-buffers
          const buf = chunk.slice()
          const part = new PushPathRequest()
          part.setChunk(buf as Buffer)
          client.send(part)
```

carsonfarmer:
LGTM!

PR #475, opened 2020-09-03
Pow

asutula:
Integrating powergate grpc calls that are supported by Hub. Still needs some integration tests, but let's get some eyes on this and get a pre-release out.

### DIFF
--- a/packages/pow/README.md
+++ b/packages/pow/README.md
@@ -1,0 +1,19 @@
+# @textile/buckets
+
+**Info**
+
+You should use the `@textile/hub` library to integrate pow into your application.
+
+**Install**
+
+```js
+npm install --save git@github.com:textileio/js-hub.git
+```
+
+**Usage**
+
+```js
+import { Buckets } from '@textile/hub'
+```
+
+[Read the docs](https://textileio.github.io/js-hub/).
--- a/packages/pow/README.md
+++ b/packages/pow/README.md
@@ -13,7 +13,7 @@ npm install --save git@github.com:textileio/js-hub.git
 **Usage**
 
 ```js
-import { Buckets } from '@textile/hub'
+import { Pow } from '@textile/hub'
 ```
 
 [Read the docs](https://textileio.github.io/js-hub/).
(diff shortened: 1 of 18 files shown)


andrewxhill:
hey @asutula is there something simple blocking here?

andrewxhill:
we need to do a draft release so staging is usable, wondering if we add this in

packages/pow/README.md
```diff
@@ -0,0 +1,19 @@
...
+**Install**
+
+```js
+npm install --save git@github.com:textileio/js-hub.git
+```
+
+**Usage**
+
+```js
+import { Buckets } from '@textile/hub'
```
andrewxhill, line 16: copy/pasta

packages/pow/src/pow.ts
```diff
@@ -0,0 +1,237 @@
...
+  newAddr,
+  peers,
+  sendFil,
+  show,
+  showAll,
+} from './api'
+
+const logger = log.getLogger('users')
+
+export class Pow extends GrpcAuthentication {
```
andrewxhill, line 41: I think at minimum all the public methods need a doc string. ideally at least 1 or 2 examples for the top level Pow class for now (i'll open a ticket for creating an example for the other public methods later)

You can see an example of what this looks like here,

https://textileio.github.io/js-hub/docs/hub.buckets

The Buckets class is documented with 3 examples, those feel like an example or a few good minimum to give someone an idea of how to get started. 

The table below, "Methods" shows the list of all public methods with the one line description of the method, that's why i think those one-liners are a minimum for each of the new methods

andrewxhill (approved):

packages/pow/README.md
asutula: done

packages/pow/src/pow.ts
asutula: Ok got the descriptions and params in for the public methods. Will address https://github.com/textileio/js-hub/issues/489 later.

PR #479, opened 2020-09-04
Expose date param for specifying auth timeout

### LINKED ISSUE #478: Expose method for apps to increase auth timeout
It doesn't look easily possible in withKeyInfo at least, possibly other helper methods.

cc @carsonfarmer in case you see a quick fix

carsonfarmer:

## Description

This exposes the date parameter for the underlying Context's auth timeout directly in `withKeyInfo`. The parameter was added as a 3rd optional parameter to avoid breaking changes, though ideally in the future it would be moved forward or combined into an options arg.

Fixes #478. It would also benefit from https://github.com/textileio/js-threads/pull/525, though does not require it.

## Type of change

- [ ] This change requires a documentation update

## How Has This Been Tested?

Existing tests should continue to pass.

## Checklist:

### DIFF
--- a/packages/grpc-authentication/src/index.ts
+++ b/packages/grpc-authentication/src/index.ts
@@ -91,9 +91,9 @@ export class GrpcAuthentication extends GrpcConnection {
    * }
    * ```
    */
-  static async withKeyInfo(key: KeyInfo, host = defaultHost, debug = false) {
+  static async withKeyInfo(key: KeyInfo, host = defaultHost, debug = false, date?: Date) {
     const context = new Context(host)
-    await context.withKeyInfo(key)
+    await context.withKeyInfo(key, date)
     return new GrpcAuthentication(context, debug)
   }
 


packages/grpc-authentication/src/index.ts
```diff
@@ -91,9 +91,9 @@ export class GrpcAuthentication extends GrpcConnection {
    * }
    * ```
    */
-  static async withKeyInfo(key: KeyInfo, host = defaultHost, debug = false) {
+  static async withKeyInfo(key: KeyInfo, host = defaultHost, debug = false, date?: Date) {
```
andrewxhill, line 94: maybe it's time to move this to `withKeyInfo(key, options)`? that options object is used all over the place.

andrewxhill: alternatively, we could move `host` last...

carsonfarmer: Yes I'd like to use an options arg... I was just sneaking this in to avoid any API changes. But if we're ok with some API breakage here, then let's go with the options arg...

andrewxhill: oh, that's why i was mentioning. now is the moment since the next release already contains them

PR #514, opened 2020-09-09
fix(docs): updates a few missing types from buckets/threads

andrewxhill:


etc/hub.api.md
```diff
@@ -909,13 +908,14 @@ export const Where: typeof Criterion;
 
 // @public
 export interface WithKeyInfoOptions extends WithUserAuthOptions {
-    // (undocumented)
     date?: Date;
+    debug?: boolean;
+    host?: string;
```
deefactorial, line 913: Isn't the `host` and `debug` properties implied if it extends the `WithUserAuthOptions` type with those properties in it? Why the nested extensions if they are duplicated?

packages/grpc-authentication/src/index.ts
```diff
@@ -8,20 +8,41 @@ import { Client } from '@textile/hub-threads-client'
...
+   */
+  debug?: boolean
+  /**
+   * Optional. Future date for the session to expire. Short time periods are strongly recommended. (Default 30 minutes).
+   */
   date?: Date
+  /**
+   * Development purposes only.
+   */
+  host?: string
```
deefactorial, line 22: same here with `host` and `debug` properties being implied

deefactorial (commented):
Questions

etc/hub.api.md
andrewxhill: it's all about the static docs generation. unfortunately, our generator doesn't pick up the implied properties.

packages/grpc-authentication/src/index.ts
andrewxhill: same yeah

carsonfarmer (approved):
Ét voila!

PR #543, opened 2020-09-16
Use older version of pow grpc lib

asutula:
It was using the latest version which has some breaking changes over the version of powergate that is actually running in staging (0.5.0-rc1). Using the 0.5.0-rc1 version of the grpc lib fixes the problem.


carsonfarmer (approved):
Good catch @asutula1

andrewxhill:
closing now, we've moved beyond

PR #549, opened 2020-09-17
feat(ci): adds minimal api connector for ci action

andrewxhill:
removed this in a previous release. spaced it on why it was needed, added a note in the docstring now


carsonfarmer (approved):
Sounds good to me!

PR #550, opened 2020-09-17
Update grpc releases + sidebar updates + wrapper types updates

carsonfarmer:

## Description

Very minor update to include the updated method name (abort -> discard). All other API changes are not exposed.


andrewxhill (approved):

packages/buckets/src/api/index.ts
```diff
@@ -80,23 +79,37 @@ export type RootObject = {
...
  */
-export type ListPathItemObject = {
+export type PathItemObject = {
   cid: string
   name: string
   path: string
   size: number
   isDir: boolean
-  itemsList: Array<ListPathItemObject>
+  itemsList: Array<PathItemObject>
```
andrewxhill, line 103: i think this should just be `items` now too

packages/buckets/src/api/index.ts
```diff
@@ -167,10 +180,12 @@ const convertPathItem = (item: ListPathItem): ListPathItemObject => {
     size: item.getSize(),
     isDir: item.getIsDir(),
     itemsList: list ? list.map(convertPathItem) : [],
+    itemsCount: item.getItemsCount(),
```
andrewxhill, line 183: same, 'items'

andrewxhill: oop, wrong line. itemslist => item

packages/buckets/src/api/index.ts
```diff
@@ -80,23 +79,37 @@ export type RootObject = {
...
+export type PathItemObject = {
   cid: string
   name: string
   path: string
   size: number
   isDir: boolean
-  itemsList: Array<ListPathItemObject>
+  itemsList: Array<PathItemObject>
+  itemsCount: number
+  metadata?: Metadata
```
andrewxhill, line 105: these metadata are looking a bit ugly to use the client. can we retype them to just make them easy. 

```
const convertMetadata = (metadata?: Metadata): PathMetadata | undefined => {
  if (!metadata) return
  const roles = metadata.getRolesMap()
  const typedRoles = new Map()
  roles.forEach((entry, key) => typedRoles.set(key, entry))
  const response: PathMetadata = {
    updatedAt: metadata.getUpdatedAt(),
    roles: typedRoles,
  }

  return response
}
```

andrewxhill: or maybe you have something nicer there. it's just the `PathAccessRoleMap` doesn't appear to act like a regular map created from scratch. you get some type mismatch :/ maybe not a huge deal. take a look though

andrewxhill: oop, i see. you gave that a nice type too. so maybe okay!!

packages/buckets/src/index.ts
```diff
@@ -1,25 +1,24 @@
...
   ArchiveResponse,
   ArchiveStatusResponse,
   ArchiveWatchResponse,
   CreateResponse,
-  ListPathItem,
+  LinksResponse,
+  ListIpfsPathResponse,
   ListPathResponse,
   ListResponse,
+  PathItem,
```
andrewxhill, line 12: chuck in the PathMetadata

andrewxhill (approved):

packages/buckets/src/index.ts
andrewxhill: oop, i meant MetadataObject

carsonfarmer: gotcha

carsonfarmer: metadata object is already exported from ./api

carsonfarmer: and metadata (singular) comes from buckets_pb, so I think we're good on both fronts

andrewxhill: will `import { MetadataObject } from "@textile/hub"` work?

andrewxhill: ahhhh, sorry, my bad. of course

andrewxhill: reading the wrong lines. merge away

PR #552, opened 2020-09-17
feat(api): adds push path access and pull path access

andrewxhill:


carsonfarmer (approved):
oooooh, nice!

carsonfarmer:
Would be nice to port the tests for this over as well @andrewxhill , but maybe we do that in a subsequent PR.

andrewxhill:
what are testts? 

okay, good point. will add a couple

packages/buckets/src/buckets.spec.ts
```diff
@@ -19,9 +21,18 @@ describe('Buckets...', () => {
   let buck: CreateObject
   let fileSize: number
 
+  let dev: SignupResponse.AsObject
+  const apiKeyInfo = { key: '' }
+
   before(async () => {
```
carsonfarmer, line 27: Note for future reference, we shouldn't be using arrow functions for async functions with mocha anymore (I'm also guilty of this), the behavior is often unexpected, because we end up with the wrong "this" context. Any `before`, `after`, `beforeEach`, `afterEach`, and `it`'s should all be `async function () {...}`, whereas `describe` and `context` can be arrow functions (but should probably just be `async function` to be consistent and safe.

packages/buckets/src/buckets.spec.ts
```diff
@@ -242,4 +253,97 @@ describe('Buckets...', () => {
...
+    const alice = PrivateKey.fromRandom()
+    let aliceBuckets: Buckets | undefined
+    let bobBuckets: Buckets | undefined
+    let aliceThread: string | undefined
+    let rootKey: string | undefined
+    const sharedPath = 'path/to'
+    const sharedFile = 'path/to/file2.jpg'
+    const privatePath = 'dir1/file1.jpg'
+    const pth = path.join(__dirname, '../../..', 'testdata')
+    before(async function () {
```
carsonfarmer, line 269: Nice!

packages/buckets/src/buckets.spec.ts
carsonfarmer: Looks like you're already doing this with new tests, so kudos!

packages/buckets/src/buckets.spec.ts
```diff
@@ -242,4 +253,97 @@ describe('Buckets...', () => {
...
+      if (isBrowser) return this.skip()
+      if (!bobBuckets || !aliceBuckets || !aliceThread || !rootKey) throw Error('setup failed')
+
+      try {
+        const stream = fs.createReadStream(path.join(pth, 'file2.jpg'))
+        await bobBuckets.pushPath(rootKey, 'path/to/bobby.jpg', stream)
+        throw wrongError
+      } catch (err) {
+        expect(err).to.not.equal(wrongError)
+        expect(err.message).to.equal('2')
```
carsonfarmer, line 345: Is it possible we're just not passing the full error message along from pushPath?

carsonfarmer (approved):
There is one "nit" comment, and a few others that don't really need addressing. LGTM!

packages/buckets/src/buckets.spec.ts
carsonfarmer: Yeh that was the case:
https://github.com/textileio/js-hub/blob/8834c8ffee622ca4d4ab0fc6c8ca072c56249b3d/packages/buckets/src/api/index.ts#L426. We're swallowing the error message and only passing along the error code there.

carsonfarmer: What we really want is something more like this:

```
onEnd: async (status: grpc.Code, message: string, _trailers: grpc.Metadata) => {
    if (status !== grpc.Code.OK) {
      reject(new Error(message))
    }
  resolve()
},
```

carsonfarmer: We should actually just include that fix this in PR to save ourselves the future headache, and to fix that test.

carsonfarmer (changes requested):
See note about extraneous console.log, and the case of the missing error message :)

packages/buckets/src/buckets.spec.ts
andrewxhill: done

packages/buckets/src/buckets.spec.ts
andrewxhill: okay, went ahead and took care of the others while in there

andrewxhill:
stuck with a mystery. can't get this to fail local or success on CI https://github.com/textileio/js-hub/pull/552/checks?check_run_id=1146111867#step:8:95

carsonfarmer (approved):

carsonfarmer:
Thanks for these updates @andrewxhill, looks great!

PR #580, opened 2020-09-25
fix(docs): adds missing archive types to sidebar

andrewxhill:


carsonfarmer (approved):
You wanna get this updated before release as well? LGTM!

PR #581, opened 2020-09-25
feat(buckets): adds bootstrap cid

andrewxhill:


carsonfarmer (approved):
This is great, let's get this in before the release!

PR #589, opened 2020-09-28
Add ArchiveConfig API

asutula:
Adds client API for getting the default Bucket `ArchiveConfig`, setting a new default `ArchiveConfig`, and providing a custom `ArchiveConfig` to any call to `archive()`.

Here's what an `ArchiveConfig` looks like (it's basically a Powergate FFS `FilConfig`):
```
export interface ArchiveConfig {
  /**
   * RepFactor indicates the desired amount of active deals
   * with different miners to store the data. While making deals
   * the other attributes of FilConfig are considered for miner selection.
   */
  repFactor: number
  /**
   * DealMinDuration indicates the duration to be used when making new deals.
   */
  dealMinDuration: number
  /**
   * ExcludedMiners is a set of miner addresses won't be ever be selected
   *when making new deals, even if they comply to other filters.
   */
  excludedMiners: Array<string>
  /**
   * TrustedMiners is a set of miner addresses which will be forcibly used
   * when making new deals. An empty/nil list disables this feature.
   */
  trustedMiners: Array<string>
  /**
   * CountryCodes indicates that new deals should select miners on specific countries.
   */
  countryCodes: Array<string>
  /**
   * Renew indicates deal-renewal configuration.
   */
  renew?: ArchiveRenew
  /**
   * Addr is the wallet address used to store the data in filecoin
   */
  addr: string
  /**
   * MaxPrice is the maximum price that will be spent to store the data, 0 is no max
   */
  maxPrice: number
  /**
   *
   * FastRetrieval indicates that created deals should enable the
   * fast retrieval feature.
   */
  fastRetrieval: boolean
  /**
   * DealStartOffset indicates how many epochs in the future impose a
   * deadline to new deals being active on-chain. This value might influence
   * if miners accept deals, since they should seal fast enough to satisfy
   * this constraint.
   */
  dealStartOffset: number
}

/**
 * ArchiveRenew contains renew configuration for a ArchiveConfig.
 */
export interface ArchiveRenew {
  /**
   * Enabled indicates that deal-renewal is enabled for this Cid.
   */
  enabled: boolean
  /**
   * Threshold indicates how many epochs before expiring should trigger
   * deal renewal. e.g: 100 epoch before expiring.
   */
  threshold: number
}
```

### DIFF
--- a/etc/hub.api.md
+++ b/etc/hub.api.md
@@ -24,9 +24,8 @@ import { GetThreadRequest } from '@textile/users-grpc/users_pb';
 import { GetThreadResponse } from '@textile/users-grpc/users_pb';
 import { grpc } from '@improbable-eng/grpc-web';
 import { GrpcConnection } from '@textile/grpc-connection';
-import { Identity as Identity_2 } from '@textile/threads-core';
 import { InfoResponse } from '@textile/grpc-powergate-client/dist/ffs/rpc/rpc_pb';
-import { Libp2pCryptoIdentity } from '@textile/threads-core';
+import { JSONSchema3or4 } from 'to-json-schema';
 import { LinksResponse } from '@textile/buckets-grpc/buckets_pb';
 import { ListDealRecordsConfig } from '@textile/grpc-powergate-client/dist/ffs/rpc/rpc_pb';
 import { ListInboxMessagesRequest } from '@textile/users-grpc/users_pb';
@@ -84,6 +83,20 @@ export type APISig = {
 };
 
 // @public
+export interface ArchiveConfig {
+    addr: string;
+    countryCodes: Array<string>;
+    dealMinDuration: number;
+    dealStartOffset: number;
+    excludedMiners: Array<string>;
+    fastRetrieval: boolean;
+    maxPrice: number;
+    renew?: ArchiveRenew;
+    repFactor: number;
+    trustedMiners: Array<string>;
+}
+
+// @public
 export type ArchiveDealInfo = {
     proposalCid: string;
     miner: string;
@@ -97,6 +110,17 @@ export type ArchiveInfo = {
 };
 
 export { ArchiveInfoResponse }
+
+// @public
+export interface ArchiveOptions {
+    archiveConfig?: ArchiveConfig;
+}
+
+// @public
+export interface ArchiveRenew {
+    enabled: boolean;
+    threshold: number;
+}
 
 export { ArchiveResponse }
 
@@ -116,7 +140,7 @@ export { ArchiveWatchResponse }
 // @public
 export class Buckets extends GrpcAuthentication {
     // @beta
-    archive(key: string): Promise<void>;
+    archive(key: string, options?: ArchiveOptions): Promise<void>;
     // @beta
     archiveInfo(key: string): Promise<ArchiveInfo>;
     // @beta
@@ -128,6 +152,8 @@ export class Buckets extends GrpcAuthentication {
     }, err?: Error) => void): Promise<() => void>;
     static copyAuth(auth: GrpcAuthentication, options?: CopyAuthOptions): Buckets;
     create(name: string, isPrivate?: boolean, cid?: string): Promise<CreateObject>;
+    // @beta
+    defaultArchiveConfig(key: string): Promise<ArchiveConfig>;
     getOrCreate(name: string, threadName?: string, isPrivate?: boolean, cid?: string, threadID?: string): Promise<{
         root?: RootObject;
         threadID?: string;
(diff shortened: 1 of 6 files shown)


jimmylee:
This is perfect and more than I was expecting! Thank you

etc/hub.api.md
```diff
@@ -289,63 +327,49 @@ export class Client {
     delete(threadID: ThreadID, collectionName: string, IDs: string[]): Promise<void>;
     deleteCollection(threadID: ThreadID, name: string): Promise<void>;
     deleteDB(threadID: ThreadID): Promise<void>;
-    find<T = any>(threadID: ThreadID, collectionName: string, query: QueryJSON): Promise<InstanceList<T>>;
-    findByID<T = any>(threadID: ThreadID, collectionName: string, ID: string): Promise<Instance<T>>;
+    find<T = unknown>(threadID: ThreadID, collectionName: string, query: QueryJSON): Promise<T[]>;
```
asutula, line 330: Lots of changes in here unrelated to my updates, not sure what's up with that.

packages/buckets/package.json
```diff
@@ -30,7 +30,7 @@
   "license": "MIT",
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.13.0",
-    "@textile/buckets-grpc": "2.0.1",
+    "@textile/buckets-grpc": "2.1.0",
```
asutula, line 33: New textile release that has the bucket archive grpc updates

packages/buckets/src/api/index.ts
```diff
@@ -118,6 +123,118 @@ export type PathObject = {
   root?: RootObject
 }
 
+/**
+ * ArchiveConfig is the desired state of a Cid in the Filecoin network.
+ */
+export interface ArchiveConfig {
+  /**
+   * RepFactor (ignored in Filecoin testnet) indicates the desired amount of active deals
```
asutula, line 131: Marking some of these as ignored in the comments, but want to keep them in the code for forward compatibility.

asutula:
Not super confident in my Lerna knowledge... let me know if there are any things that need to be done to prep for release.

etc/hub.api.md
carsonfarmer: Huh, they must not have updated in my PR?

carsonfarmer: All good though, these are correct.

andrewxhill (approved):
sick!

packages/buckets/src/api/index.ts
carsonfarmer: But you don't wanna mark them as optionals?

packages/buckets/src/api/index.ts
```diff
@@ -118,6 +123,118 @@ export type PathObject = {
...
+   */
+  enabled: boolean
+  /**
+   * Threshold indicates how many epochs before expiring should trigger
+   * deal renewal. e.g: 100 epoch before expiring.
+   */
+  threshold: number
+}
+
+const fromProtoArchiveConfig = (protoConfig: ProtoArchiveConfig): ArchiveConfig => {
```
carsonfarmer, line 196: Niiice

packages/buckets/src/api/index.ts
```diff
@@ -658,15 +775,65 @@ export async function bucketsPullPathAccessRoles(
...
  * archive creates a Filecoin bucket archive via Powergate.
  * @internal
  * @param key Unique (IPNS compatible) identifier key for a bucket.
+ * @param options Options that control the behavior of the bucket archive
  */
-export async function bucketsArchive(api: GrpcConnection, key: string, ctx?: ContextInterface) {
+export async function bucketsArchive(
+  api: GrpcConnection,
+  key: string,
+  options?: ArchiveOptions,
```
carsonfarmer, line 828: This is a pretty important breaking change here, especially just because they both take objects as args. I think it most cases, we're not using ctx directly, so it should be pretty safe. But we'll have to be sure this works with slates usage.

carsonfarmer (approved):
Love it @asutula, easy one to review, the docs are great!

packages/buckets/src/api/index.ts
asutula: No, because if someone is going to set a new default, they should set it to what they will want in the future, or be explicit about providing a value like 0 (which will not work in the future, but at least they know they set it that way)

packages/buckets/src/api/index.ts
asutula: Yea, I don't even understand context. I never see it used anywhere within the sdk, but it looks to just be available internally. All public apis don't allow it to be passed in as far as I can tell.

andrewxhill: just a note. this shouldn't be a breaking change. we don't promote the use of these internal functions. on the public api, `Buckets.archive` options was just added as an optional second param, so non-breaking there too

asutula: Great, thanks for the confirmation.

PR #598, opened 2020-09-30
fix(api): larger file block size in bucket push

andrewxhill:


carsonfarmer (approved):
Pushed a minor nit update that fixes the test for permission denied. This should be GTG now!

PR #612, opened 2020-10-01
fix(lint): rm console log; add specific buffer test

andrewxhill:
i was just doing this chore https://github.com/textileio/js-examples/pull/37 and hit something really strange in the React Native example, where it errors out on the file chunking method. it says `... not an interator` and crashes. so chucked this test in here to try and see if it existed elsewhere. but haven't found it yet. will return after next release.


carsonfarmer (approved):
Let's get this in before we do a release. Good to have the console.log outta there.

PR #618, opened 2020-10-08
Adds abort controller support + root fix

carsonfarmer:

## Description

This PR adds two new "features", both of which are optional. This should avoid and major breaking changes, but should provide improvements in behavior when pushing buckets in most cases. Note that eventually, some of these options should/will be made non-optional/the default.

This adds support for an AbortSignal from an AbortController to be used to ensure that a given pushPath connection with a remote buckd service is closed. It essentially adds support for canceling an ongoing Bucket pushPath call. This should be used by callers at all times when there is the possibility that a pushPath will timeout, error, or otherwise not complete safely. So in practice, an AbortController should be used all the time. The nice thing here is that you can use the same signal for multiple calls, so that in the case of _any_ failure, all ongoing transactions can be aborted.

This also adds better default behavior for specifying the current "root" of a bucket during a pushPath. This ensures that only fast-forward only pushes are possible, and should now error properly if a caller tries to pushPath to a "stale" bucket state. The default here is to simply fire off another listPath call to get the latest bucket root. If a caller would like to avoid an additional API call, they should query the root themselves, maintain a reference to this, and then update it accordingly. Some additional examples could be added to help illustrate this. For instance, in the event of multiple pushes in a row, the "root" can be passed from one push to the next.

This fixes a number of issues we've been observing with buckets lately, though the full extend of the fix may reach beyond this. Additionally, the AbortController fix only helps if it is used. We'll have to create some good docs to support/and promote its use.

## Type of change

## How Has This Been Tested?

All local tests should continue to pass (should be backward compatible), and new tests have been created to reflect the new features/options. Note that some regression tests should be created in the future, once these options become the default.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,17 +5,8 @@
     "commonjs": true,
     "node": true
   },
-  "extends": [
-    "plugin:prettier/recommended",
-    "plugin:@typescript-eslint/recommended",
-    "prettier/@typescript-eslint",
-    "plugin:import/typescript"
-  ],
-  "plugins": [
-    "@typescript-eslint",
-    "prettier",
-    "import"
-  ],
+  "extends": ["plugin:prettier/recommended", "plugin:@typescript-eslint/recommended", "prettier/@typescript-eslint"],
+  "plugins": ["@typescript-eslint", "prettier"],
   "globals": {
     "Atomics": "readonly",
     "SharedArrayBuffer": "readonly"
@@ -30,17 +21,8 @@
     "@typescript-eslint/explicit-function-return-type": 0,
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/explicit-member-accessibility": 0,
-    "@typescript-eslint/no-unused-vars": [0, {"argsIgnorePattern": "^_"}],
+    "@typescript-eslint/no-unused-vars": [0, { "argsIgnorePattern": "^_" }],
     "no-undefined": 0,
-    "default-case": 0,
-    "import/no-absolute-path": 2,
-    "import/no-cycle": 1,
-    "import/export": 2,
-    "import/no-named-as-default": 2,
-    "import/no-named-as-default-member": 2,
-    "import/first": 2,
-    "import/order": 2,
-    "import/newline-after-import": 2,
-    "import/no-default-export": 2
+    "default-case": 0
   }
 }
--- a/packages/buckets/package.json
+++ b/packages/buckets/package.json
@@ -41,6 +41,7 @@
     "@textile/security": "^0.3.0",
     "@textile/threads-id": "^0.1.15",
     "@types/next-tick": "^1.0.0",
+    "abort-controller": "^3.0.0",
     "cids": "^0.8.0",
     "event-iterator": "^2.0.0",
     "it-block": "^2.0.0",
--- a/packages/buckets/src/api/index.ts
+++ b/packages/buckets/src/api/index.ts
@@ -44,6 +44,7 @@ import { APIService, APIServicePushPath } from '@textile/buckets-grpc/buckets_pb
 import { Context, ContextInterface } from '@textile/context'
 import { GrpcConnection } from '@textile/grpc-connection'
 import { WebsocketTransport } from '@textile/grpc-transport'
+import type { AbortSignal } from 'abort-controller'
 import CID from 'cids'
 import { EventIterator } from 'event-iterator'
 import log from 'loglevel'
@@ -52,8 +53,31 @@ import { File, normaliseInput } from './normalize'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const block = require('it-block')
-
 const logger = log.getLogger('buckets-api')
+
+/**
+ * PushOptions provides additional options for controlling a push to a bucket path.
+ */
+export interface PushOptions {
+  /**
+   * A callback function to use for monitoring push progress.
+   */
+  progress?: (num?: number) => void
+  /**
+   * The bucket root path as a string, or root object. Important to set this property when
+   * there is a possibility of multiple parallel pushes to a bucket. Specifying this property
+   * will enforce fast-forward only updates. It not provided explicitly, the root path will
+   * be fetched via an additional API call before each push.
+   */
+  root?: RootObject | string
+
+  /**
+   * An optional abort signal to allow cancelation or aborting a bucket push.
+   */
+  signal?: AbortSignal
+}
+
+export const AbortError = new Error('aborted')
 
 /**
  * The expected result format from pushing a path to a bucket
@@ -491,7 +515,7 @@ export async function bucketsListIpfsPath(
  * @param key Unique (IPNS compatible) identifier key for a bucket.
  * @param path A file/object (sub)-path within a bucket.
  * @param input The input file/stream/object.
- * @param opts Options to control response stream. Currently only supports a progress function.
+ * @param opts Options to control response stream.
  * @remarks
  * This will return the resolved path and the bucket's new root path.
  * @example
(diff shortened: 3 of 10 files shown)


.eslintrc.json
```diff
@@ -5,17 +5,8 @@
...
-    "plugin:@typescript-eslint/recommended",
-    "prettier/@typescript-eslint",
-    "plugin:import/typescript"
-  ],
-  "plugins": [
-    "@typescript-eslint",
-    "prettier",
-    "import"
-  ],
+  "extends": ["plugin:prettier/recommended", "plugin:@typescript-eslint/recommended", "prettier/@typescript-eslint"],
```
carsonfarmer, line 8: All those linter settings were really annoying me, so I'm sneaking this change in as well.

.eslintrc.json
```diff
@@ -30,17 +21,8 @@
     "@typescript-eslint/explicit-function-return-type": 0,
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/explicit-member-accessibility": 0,
-    "@typescript-eslint/no-unused-vars": [0, {"argsIgnorePattern": "^_"}],
+    "@typescript-eslint/no-unused-vars": [0, { "argsIgnorePattern": "^_" }],
     "no-undefined": 0,
-    "default-case": 0,
-    "import/no-absolute-path": 2,
```
carsonfarmer, line 36: Removing all the import ones in favor of letting prettier/eslint take care of this. This _might_ lead to some import reordering in future PRs... but so be it!

packages/buckets/package.json
```diff
@@ -41,6 +41,7 @@
     "@textile/security": "^0.3.0",
     "@textile/threads-id": "^0.1.15",
     "@types/next-tick": "^1.0.0",
+    "abort-controller": "^3.0.0",
```
carsonfarmer, line 44: To support abort controller APIs on all platforms (node, browser, etc)

packages/buckets/src/api/index.ts
```diff
@@ -52,9 +53,32 @@ import { File, normaliseInput } from './normalize'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const block = require('it-block')
-
 const logger = log.getLogger('buckets-api')
 
+/**
```
carsonfarmer, line 58: Pulled this out into its own interface, before it was just the progress thing, now we have root and signal.

packages/buckets/src/api/index.ts
```diff
@@ -512,21 +536,33 @@ export async function bucketsPushPath(
...
   return new Promise<PushPathResult>(async (resolve, reject) => {
-    // Only process the first  input if there are more than one
+    // Only process the first input if there are more than one
     const source: File | undefined = (await normaliseInput(input).next()).value
     const client = grpc.client<PushPathRequest, PushPathResponse, APIServicePushPath>(APIService.PushPath, {
       host: api.serviceHost,
       transport: api.rpcOptions.transport,
       debug: api.rpcOptions.debug,
     })
+    // Send a close event to the bucket api upon abort
```
carsonfarmer, line 550: So, we listen to the abort event on the signal, and simply close the connection and reject the promise with an AbortError.

packages/buckets/src/api/index.ts
```diff
@@ -512,21 +536,33 @@ export async function bucketsPushPath(
...
+    if (opts?.signal !== undefined) {
+      opts.signal.addEventListener('abort', () => {
+        client.close()
+        return reject(AbortError)
+      })
+    }
     client.onMessage((message) => {
+      // Let's just make sure we haven't aborted this outside this function
```
carsonfarmer, line 558: We also want to check each time we receive a new message from the remote, if we've aborted this locally. If yes, close the connection right away and reject.

packages/buckets/src/api/index.ts
```diff
@@ -542,42 +578,63 @@ export async function bucketsPushPath(
...
       } else {
-        reject(new Error('Invalid reply'))
+        return reject(new Error('Invalid reply'))
       }
     })
     client.onEnd((code, msg) => {
       if (code !== grpc.Code.OK) {
         const message = msg ? msg : code.toString()
-        reject(new Error(message))
+        return reject(new Error(message))
```
carsonfarmer, line 592: These returns avoid having to deal with catching already closed connections later on down the line. Once we close a connection, we're never going to use it again, so we can just return here.

packages/buckets/src/api/index.ts
```diff
@@ -542,42 +578,63 @@ export async function bucketsPushPath(
...
+        return resolve()
       }
     })
     if (source) {
       const head = new PushPathRequest.Header()
       head.setPath(source.path || path)
       head.setKey(key)
+      // Setting root here ensures pushes will error if root is out of date
+      let root = ''
+      if (opts?.root) {
```
carsonfarmer, line 603: The new behavior is that we "trust" the caller if they explicitly provide a root, but otherwise, make another API call to check the current known root.

packages/buckets/src/api/index.ts
```diff
@@ -542,42 +578,63 @@ export async function bucketsPushPath(
...
+      } else {
+        // Otherwise, make a call to list path to get the latest known root
+        const head = await bucketsListPath(api, key, '', ctx)
+        root = head.root?.path ?? '' // Shouldn't ever be undefined here
+      }
+      head.setRoot(root)
       const req = new PushPathRequest()
       req.setHeader(head)
       const metadata = { ...api.context.toJSON(), ...ctx?.toJSON() }
+      // Let's just make sure we haven't aborted this outside this function
```
carsonfarmer, line 615: Just in case, we might as well just reject on not even start the thing if we've already aborted for some reason.

carsonfarmer: This is useful in cases where we might have a bunch of pushes all queued up.

packages/buckets/src/api/index.ts
```diff
@@ -542,42 +578,63 @@ export async function bucketsPushPath(
...
+      if (opts?.signal?.aborted) {
+        return reject(AbortError)
+      }
       client.start(metadata)
       client.send(req)
 
       if (source.content) {
         const process = await block({ size: 256000, noPad: true })
         for await (const chunk of process(source.content)) {
+          // Let's just make sure we haven't aborted this outside this function
```
carsonfarmer, line 625: Again, for safety reasons, each new chuck we're about to push is a good opportunity to check if we've aborted the call.

packages/buckets/src/api/index.ts
```diff
@@ -542,42 +578,63 @@ export async function bucketsPushPath(
...
+            return reject(AbortError)
+          }
           const buf = chunk.slice()
           const part = new PushPathRequest()
           part.setChunk(buf as Buffer)
           client.send(part)
         }
-        client.finishSend()
       }
+      // We only need to finish send here if we actually started
```
carsonfarmer, line 636: This was in the wrong place before, and would have left open connections if we _never_ send any data at all.

packages/buckets/src/buckets.spec.ts
```diff
@@ -271,6 +273,107 @@ describe('Buckets...', function () {
       }
     })
   })
+
+  describe('multiple pushes', function () {
+    it('should throw if roots are out of sync/order', async function () {
```
carsonfarmer, line 278: This is just an additional test on using root directly. We want to make sure we get an error when we expect to get one.

packages/buckets/src/buckets.spec.ts
```diff
@@ -271,6 +273,107 @@ describe('Buckets...', function () {
...
+        }) // This should never fire
+        // Because we should jump down to this first
+        .catch((err) => expect(err.message).to.include('non-fast-forward'))
+
+      // Confirm that our second-level dir only includes one of them
+      const { item } = await client.listPath(rootKey, '')
+      expect(item?.items[1].items).to.have.length(1)
+    })
+
+    it.skip('should not be able to push in parallel', async function () {
```
carsonfarmer, line 309: This isn't in effect yet, but it would be super useful to have this built into buckets. I know @sanderpick already mentioned he's like to add something like this, so I have a test here ready to go.

packages/buckets/src/buckets.spec.ts
```diff
@@ -271,6 +273,107 @@ describe('Buckets...', function () {
...
+        expect(err.message).to.include('non-fast-forward')
+      }
+      // Confirm that our second-level dir only includes one of them
+      const { item } = await client.listPath(rootKey, '')
+      expect(item?.items[1].items).to.have.length(1)
+    })
+  })
+
+  describe('aborting', function () {
+    it('should allow an abort controler to signal a cancel event on a push', async function () {
```
carsonfarmer, line 344: Just making sure we can abort an inflight push. Here, we use an infinite stream of "bytes", that in theory would never close the connection with our remote bucket API, but that is also a little bit slow. If the abort thing works as expected, it'll kill the connection and when we try to list path afterwards, we should get an error. If it doesn't work, this test will hang and timeout...

carsonfarmer: There actually is no reason to not test this in a browser, I might remove that check actually...

carsonfarmer (commented):
Some comments for your review pleasure @andrewxhill and/or @sanderpick

sanderpick (approved):
Nice! I like the controller pattern. Would it be possible to add this functionality across the board to all method? Certainly other heavy methods like `setPath`, `create`, etc., would benefit, but I can't think of a reason to not add it to them all.

packages/buckets/src/api/index.ts
andrewxhill: could you add this to sidebar.js?

carsonfarmer: Can you actually set that as a "request change" thing in your review, I'm about to walk the dog and I don't wanna forget.

carsonfarmer:
Yes @sanderpick, we can indeed. I like the pattern a lot as well. I actually started an issue here to do just this (add it everywhere), but thought it'd be good to get this one out asap. What do you think?

sanderpick:
Yep, agreed.

website/sidebars.json
```diff
@@ -108,6 +108,7 @@
       "hub.buckets.listipfspath",
       "hub.buckets.open",
       "hub.buckets.pushpath",
+      "hub.buckets.pushoptions",
```
andrewxhill, line 111: nice!

PR #651, opened 2020-10-22
fix(docs): incorrect type in withThread

andrewxhill:


etc/hub.api.md
```diff
@@ -4,6 +4,7 @@
 
 ```ts
 
+import type { AbortSignal as AbortSignal_2 } from 'abort-controller';
```
carsonfarmer, line 7: why is it doing this?

carsonfarmer (approved):
Looks good. The auto-generated code seems fishy though?

etc/hub.api.md
andrewxhill: i'm guessing you didn't rebuild this file with your older abort work

carsonfarmer: I did not, but it seems strange that it is importing as AbortSignal_2? It does the same thing with ThreadId

andrewxhill: maybe it just doesn't understand the `import type` system well. but related, should AbortSignal be exported to end users actually?

carsonfarmer: We could export it yeh, its really just a import away, but easy enough to include it. Thing is, browsers already have that API, which is why we didn't include it by default. But yeh, can't hurt.

PR #716, opened 2020-11-24
fix(api): makes default api sig in buckets use options where approp.

andrewxhill:


packages/buckets/src/buckets.ts
```diff
@@ -303,10 +347,37 @@ export class Buckets extends GrpcAuthentication {
    * }
    * ```
    */
+  async getOrCreate(name: string, options?: GetOrCreateOptions): Promise<{ root?: RootObject; threadID?: string }>
```
carsonfarmer, line 350: Probably should create a type for this return here at some point. Not a blocker though.

packages/buckets/src/buckets.ts
```diff
@@ -303,10 +347,37 @@ export class Buckets extends GrpcAuthentication {
...
+      // Using new signature
+      const threadName = options.threadName && options.threadName !== '' ? options.threadName : 'buckets'
+      const encrypted = !!options.encrypted
+      return this._getOrCreate(name, threadName, encrypted, options.cid, options.threadID)
+    }
+  }
+  /**
+   * @internal
+   */
+  async _getOrCreate(
```
carsonfarmer, line 377: Could also flag this as `private` if you wanna keep it private explicitly.

carsonfarmer (approved):
Looks good Andrew. I made two or three comments, but they shouldn't hold anything up. Merge away! We'll probably want to do a major version bump on this one I guess right? In theory not actually API breaking... so your call.

PR #722, opened 2020-11-25
feature(users): adds usage api

andrewxhill:


packages/users/src/api/index.ts
```diff
@@ -73,6 +75,52 @@ export interface GetThreadResponseObj {
   id: ThreadID
 }
 
+
+export interface UsageObj {
```
sanderpick, line 79: Is the use of `Obj` in the object names a convention you guys have been using? Feels a bit cumbersome.

sanderpick (commented):
How do the tests spin up textile and the billing service?

packages/users/src/api/index.ts
```diff
@@ -17,6 +17,8 @@ import {
   ReadInboxMessageResponse,
   DeleteInboxMessageRequest,
   DeleteSentboxMessageRequest,
+  GetUsageRequest,
```
sanderpick, line 20: IMO, I would import these an some internal name, then use the nice names as the response type for the user.

packages/users/src/api/index.ts
andrewxhill: well... it was not to conflict with the Usage type that comes back from the protos. However, and given the protos we really return Usage.AsObject obj result so went with it.

packages/users/src/api/index.ts
andrewxhill: We don't make the users use those, https://github.com/textileio/js-textile/pull/722/files#diff-0c6964ee0de0d6937c3fc3d2250393adf892a2893bb921b776dfcded5f650d8aR390 though perhaps a nicer name that GetUsageResponsObj is still in order

sanderpick: Right, I was just suggesting do something like `GetUsageRequest as GetUsageRequestInternal` or whatever the JS way of doing that is now days, then use the pleasant name `GetUsageRequest` as your custom type (the return type)

andrewxhill: oh nice, good call

carsonfarmer: An alternative here is to import it like this:

```
import * as pb from "@textile/users-grpc/api/usersd/pb/usersd_pb"
```

And then prefix everything with `pb.` when using them. this is how we do it in threads IIRC? Not saying that's the right way, just an option. The way @andrewxhill has it here might be better for tree-shaking?

carsonfarmer: (though of course, import "as" internal name is totally a good call)

packages/buckets/src/api/index.ts
```diff
@@ -95,23 +95,32 @@ export interface PushPathResult {
...
   name: string
   path: string
   createdAt: number
   updatedAt: number
   thread: string
 }
+/**
+ * @deprecated
+ */
+export type RootObject = Root
```
andrewxhill, line 123: Left these in as backward compatible types for now

packages/buckets/src/api/index.ts
```diff
@@ -120,32 +129,44 @@ export enum PathAccessRole {
   PATH_ACCESS_ROLE_ADMIN = 3,
 }
 
-export type MetadataObject = {
+export type BuckMetadata = {
```
andrewxhill, line 132: Required a prefix to not clash with Metadata in docs elsewhere

packages/users/src/api/index.ts
andrewxhill: "import * as" was causing issues, i think it was with api-extractor so have avoided it. updated with `Thing as _Thing` for now. Updated other cases of `ThingObj` and added backward compatibility where I saw them.

packages/buckets/src/api/index.ts
```diff
@@ -95,23 +95,32 @@ export interface PushPathResult {
...
  */
-export type LinksObject = {
+export type Links = {
   www: string
   ipns: string
   url: string
 }
 
+/**
+ * @deprecated
```
carsonfarmer, line 105: Niiice!

packages/buckets/src/api/index.ts
```diff
@@ -21,8 +21,8 @@ import {
   ListPathResponse,
   ListRequest,
   ListResponse,
-  Metadata,
-  PathItem,
+  Metadata as _Metadata,
```
carsonfarmer, line 24: Good call

packages/buckets/src/spec.util.ts
```diff
@@ -5,6 +5,7 @@ import { APIServiceClient, ServiceError } from '@textile/hub-grpc/api/hubd/pb/hu
 import axios from 'axios'
 import delay from 'delay'
 import { HMAC } from 'fast-sha256'
+// @ts-ignore
```
carsonfarmer, line 8: What's this for? That should come with types now...

packages/hub-threads-client/src/client.ts
```diff
@@ -35,11 +35,16 @@ Client.prototype.getThread = async function (name: string, ctx?: Context): Promi
...
-            msg.id = ThreadID.fromBytes(Buffer.from(msg.id as string, 'base64')).toString()
+          if (message) {
+            const res = {
+              name: message.getName(),
+              isDb: message.getIsDb(),
+              id: ThreadID.fromBytes(message.getId_asU8()).toString(),
+            }
+            resolve(res)
+          } else {
+            resolve()
```
carsonfarmer, line 46: This is fine (correct if you ask me), but if/when we update Typescript, its not going to like this :(

packages/hub-threads-client/src/client.ts
```diff
@@ -67,13 +72,15 @@ Client.prototype.listThreads = async function (ctx?: Context): Promise<ListThrea
       .then((meta) => {
         client.listThreads(req, meta, (err: ServiceError | null, message: ListThreadsResponse | null) => {
           if (err) reject(err)
-          const msg = message?.toObject()
-          if (msg) {
-            msg.listList.forEach((thread) => {
-              thread.id = ThreadID.fromBytes(Buffer.from(thread.id as string, 'base64')).toString()
-            })
+          const lst = message?.getListList()
```
carsonfarmer, line 75: These are lame, I wish we could configure that on our builds...

packages/hub-threads-client/src/client.ts
```diff
@@ -67,13 +72,15 @@ Client.prototype.listThreads = async function (ctx?: Context): Promise<ListThrea
...
+          const listList = []
+          if (lst) {
+            for (let thrd of lst) {
+              const row = thrd.toObject()
+              listList.push({...row, id: ThreadID.fromBytes(thrd.getId_asU8()).toString()})
```
carsonfarmer, line 80: nice!

packages/users/src/spec.util.ts
```diff
@@ -1,6 +1,7 @@
 import axios from 'axios'
 import delay from 'delay'
 import { HMAC } from 'fast-sha256'
+// @ts-ignore
 import multibase from 'multibase'
```
carsonfarmer, line 5: Same here? Is this causing issues?

packages/users/src/users.spec.ts
```diff
@@ -452,5 +452,30 @@ describe('Users...', () => {
...
+      this.timeout(10000)
+      const { user } = await signUp(ctx, addrGatewayUrl, sessionSecret)
+      if (user) dev = user
+    })
+    it('it should return the correct usage per user', async () => {
+      const tmp = new Context(addrApiurl).withSession(dev.session)
+      const { keyInfo } = await createKey(tmp, 'KEY_TYPE_ACCOUNT')
+      await ctx.withAPIKey(keyInfo?.key).withKeyInfo(keyInfo)
+      const user = new Users(ctx)
+      try {
```
carsonfarmer, line 470: This is so cool!

packages/users/src/users.ts
```diff
@@ -208,6 +210,22 @@ export class Users extends GrpcAuthentication {
...
+   * @example
+   * ```typescript
+   * import { Users } from "@textile/hub"
+   *
+   * async function example(users: Users) {
+   *    const usage = await users.getUsage()
+   * }
+   * ```
+   */
+  async getUsage(): Promise<GetUsageResponse> {
```
carsonfarmer, line 225: Side note: I wonder if we can show example output in these examples?

carsonfarmer (changes requested):
This looks great @andrewxhill, some really nice cleanups in there. The only "question" I have is around multibase. Why the ts-ignore? If this is causing issues, we should try to fix that. Perhaps its just a version bump that is required?

packages/hub-threads-client/src/client.ts
andrewxhill: so, it was `resolve(undefined)` before this change... what's the solution do you think?

packages/users/src/spec.util.ts
andrewxhill: yeah... types not found errors

packages/users/src/users.ts
andrewxhill: yeah, will add that as a follow-up.

packages/hub-threads-client/src/client.ts
carsonfarmer: I think what you have is good. I had to fight typescript over in js-threads, because the promise return type doesn't include void or undefined technically. So I guess we should throw in that case? But some of the responses _do_ return void... so long story short... I'm not sure.

packages/users/src/users.spec.ts
andrewxhill: ah @sanderpick to answer your question about tests, not very much. i only pull of testing the successful error here. not sure how to do the full billing enabled setup.

carsonfarmer (approved):
woop wood

.github/workflows/test.yml
```diff
@@ -26,49 +26,53 @@ jobs:
           HUB_EMAIL_SESSION_SECRET: hubsession
           HUB_LOG_DEBUG: true
         ports:
-          - "127.0.0.1:3006:3006"
```
carsonfarmer, line 29: Linter I guess?

packages/buckets/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@textile/buckets",
-	"version": "1.2.1",
+	"version": "2.0.2",
```
carsonfarmer, line 3: This was bumped as part of an off-main release process.

packages/buckets/package.json
```diff
@@ -31,28 +31,26 @@
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.13.0",
     "@textile/buckets-grpc": "2.1.7",
-    "@textile/context": "^0.8.0",
-    "@textile/crypto": "^0.1.2",
-    "@textile/grpc-authentication": "^0.3.1",
-    "@textile/grpc-connection": "^0.2.1",
-    "@textile/grpc-transport": "^0.0.6",
+    "@textile/context": "^0.9.2",
```
carsonfarmer, line 34: All our sub-packages should also be updated.

packages/buckets/src/api/index.ts
```diff
@@ -34,25 +34,28 @@ import {
...
   RemoveRequest,
-  Root,
+  Root as _Root,
   RootRequest,
   RootResponse,
   SetDefaultArchiveConfigRequest,
   SetPathRequest,
 } from '@textile/buckets-grpc/api/bucketsd/pb/bucketsd_pb'
-import { APIService, APIServicePushPath } from '@textile/buckets-grpc/api/bucketsd/pb/bucketsd_pb_service'
+import {
```
carsonfarmer, line 43: Most of these changes are due to @andrewxhill's prior work on making these types nicer for downstream callers. This work has already been approved.

packages/buckets/src/api/index.ts
```diff
@@ -120,31 +132,49 @@ export enum PathAccessRole {
...
+  root?: Root
+}
+/**
+ * @deprecated
+ */
+export type PathObject = Path
+
+export const CHUNK_SIZE = 32768
+
+export function* genChunks(value: Uint8Array, n: number) {
```
carsonfarmer, line 176: Naive chunk generator that simple chunks things into smaller bits, but includes everything if it can.

packages/buckets/src/api/index.ts
```diff
@@ -617,17 +647,20 @@ export async function bucketsPushPath(
       client.send(req)
 
       if (source.content) {
-        const process = await block({ size: 32768, noPad: true })
-        for await (const chunk of process(source.content)) {
+        for await (const chunk of source.content) {
```
carsonfarmer, line 650: Simple go through the chunked content... where chunks _might_ be extra big, or extra small.

packages/buckets/src/api/index.ts
```diff
@@ -617,17 +647,20 @@ export async function bucketsPushPath(
...
+            try {
+              client.close()
+            } catch {} // noop
             return reject(AbortError)
           }
-          const buf = chunk.slice()
-          const part = new PushPathRequest()
-          part.setChunk(buf as Buffer)
-          client.send(part)
+          // Naively chunk into chunks smaller than CHUNK_SIZE bytes
```
carsonfarmer, line 658: _Now_ we try to chunk it up into parts of roughly CHUNK_SIZE, which may or may not be exactly what we want, but should be close.

packages/buckets/src/api/index.ts
```diff
@@ -636,6 +669,119 @@ export async function bucketsPushPath(
   })
 }
 
+export async function bucketsPushPathNode(
```
carsonfarmer, line 672: Merged from #732

packages/buckets/src/api/index.ts
```diff
@@ -636,6 +669,119 @@ export async function bucketsPushPath(
...
+        if (opts?.signal?.aborted) {
+          // Let's just make sure we haven't aborted this outside this function
+          try {
+            // Should already have been handled
+            stream.cancel()
+          } catch {} // noop
+          return reject(AbortError)
+        }
+        // Naively chunk into chunks smaller than CHUNK_SIZE bytes
+        for (const chunklet of genChunks(chunk as Uint8Array, CHUNK_SIZE)) {
```
carsonfarmer, line 775: Now also includes work around the chunking work...

packages/buckets/src/buckets.spec.ts
```diff
@@ -18,10 +18,24 @@ const wrongError = new Error('wrong error!')
 const rightError = new Error('right error!')
 const sessionSecret = 'hubsession'
 
+describe('Buckets utils...', function () {
```
carsonfarmer, line 21: Simple test, should be expanded.

packages/buckets/src/buckets.spec.ts
```diff
@@ -438,9 +452,12 @@ describe('Buckets...', function () {
       await bobBuckets.getToken(bob)
       bobBuckets.withThread(aliceThread)
 
+      const { root } = await bobBuckets.listPath(rootKey, '')
       try {
+        // Note: Callers can also specify chunk size directly using highWaterMark:
+        // const stream = fs.createReadStream(path.join(pth, 'file2.jpg'), { highWaterMark: 32768 })
```
carsonfarmer, line 458: Updating test with a little reminder for the future. Though it shouldn't be required.

carsonfarmer (commented):
Here's a review for you @andrewxhill ... note that all tests pass!

packages/hub-threads-client/src/client.ts
andrewxhill: want to sneak in a no result error?

packages/hub-threads-client/src/client.ts
andrewxhill: once we hide the names from end users, they don't matter much

packages/buckets/src/api/index.ts
andrewxhill: should `n` be renamed since it is really `s` or `size` not number?

packages/hub-threads-client/src/client.ts
carsonfarmer: added

PR #723, opened 2020-11-25
Updated archives API with DealInfo

asutula:
There used to be:

- archive `status` that would tell you if your archive job was queued, processing or finished, and would throw an error if the bucket had never been archived
- archive `info` would provide information about an archive, but only after the underlying job was complete. if a bucket was never archived or if the underlying job was still running, it would throw an error.

Those old methods are replaced by one simple method:

bucket `archives` returns an object that looks like:

```
interface Archives {
  current?: Archive
  history: Array<Archive>
}
```

This method can be called no matter if the bucket has been archived in the past or not. No error, it will just have `undefined` and/or empty array values if appropriate. The latest archive is provided by `current` and a log of past archives is available in `history`. Once a new archive is initiated, it is immediately set to `current` and is updated as new information about job and deal status is available. `Archive` objects look like:

```
interface Archive {
  cid: string
  jobId: string
  status: ArchiveStatus
  aborted: boolean
  abortedMsg: string
  failureMsg: string
  createdAt: number
  dealInfoList: Array<DealInfo>
}
```

and importantly contains that `dealInfoList: Array<DealInfo>` which is updated as data is available.

For reference, `DealInfo` looks like:

```
export interface DealInfo {
  proposalCid: string
  stateId: number
  stateName: string
  miner: string
  pieceCid: string
  size: number
  pricePerEpoch: number
  startEpoch: number
  duration: number
  dealId: number
  activationEpoch: number
  message: string
}
```

This should be everything needed to have a complete and near real time view of how bucket archives are stored on filecoin.


asutula:
Got into rebase commit rewording hell trying to satisfy the commitsar, so disabled it for now. We can re-enable it after this is merged, but would really like a solution that runs pre-commit or to not do these sorts of checks unless there is a good reason.

packages/buckets/src/api/index.ts
```diff
@@ -1,16 +1,18 @@
...
   ArchiveConfig as ProtoArchiveConfig,
-  ArchiveInfoRequest,
-  ArchiveInfoResponse,
   ArchiveRenew as ProtoArchiveRenew,
   ArchiveRequest,
-  ArchiveStatusRequest,
-  ArchiveStatusResponse,
+  ArchivesRequest,
+  ArchivesResponse,
+  ArchiveStatus as PbArchiveStatus,
```
carsonfarmer, line 9: We should probably just do something consistent here across the board at some point. @andrewxhill was using `_` prefixes: https://github.com/textileio/js-textile/blob/master/packages/buckets/src/api/index.ts#L24. I also like `pb_`... no real preference though. But something consistent would be good.

packages/buckets/src/api/index.ts
```diff
@@ -1038,51 +1102,17 @@ export async function bucketsArchive(
...
-  const req = new ArchiveStatusRequest()
+export async function bucketsArchives(api: GrpcConnection, key: string, ctx?: ContextInterface): Promise<Archives> {
+  logger.debug('archives request')
+  const req = new ArchivesRequest()
   req.setKey(key)
-  const res: ArchiveStatusResponse = await api.unary(APIService.ArchiveStatus, req, ctx)
+  const res: ArchivesResponse = await api.unary(APIService.Archives, req, ctx)
+  const current = res.getCurrent()
   return {
-    key: res.getKey(),
```
carsonfarmer, line 1055: Love all the red here!

carsonfarmer (approved):
Awesome @asutula, LGTM! No real changes from my perspective. One or two "suggestions" for future improvements, but that's about it.

asutula:
Ok, I also committed changes that move to "filecoin" naming conventions instead of "pow" or "powergate":

- Renamed package directory `pow` to `hub-filecoin`
- Package `@textile/pow` is now named `@textile/hub-filecoin`
- `@textile/hub-filecoin` exports the `Filecoin` class
- `@textile/hub-filecoin` is exported from `@textile/hub` as `Filecoin`
- Code snippets/docs updated to reflect all this

carsonfarmer:
Cool with the name updates! Why prefix with `hub`? Just in case we want to have a future `@textile/filecoin` package?

carsonfarmer (approved):
Looking thru these new changes, it looks to be just name changes, yes? Also, this is a great opportunity to do some "marketing":

> Textile "pow" Javascript library is now "hub-filecoin". This new version and name change reflects ...

asutula:
> Why prefix with hub? Just in case we want to have a future @textile/filecoin package?

Was on the fence about that. My thought was just that it would be confusing if you were looking at all the Textile npm packages trying to figure out which one does what. We already have the powergate client which is clearly named, and if we have other filecoin related packages in the future, they should be clearly named as well. `@textile/filecoin` is not clear to me.

PR #725, opened 2020-11-26
feat(node): push path with duplex stream

andrewxhill:


packages/buckets/src/buckets.ts
```diff
@@ -617,6 +619,9 @@ export class Buckets extends GrpcAuthentication {
    * ```
    */
   async pushPath(key: string, path: string, input: any, options?: PushOptions): Promise<PushPathResult> {
+    if (isNode) {
+      return bucketsPushPathNode(this, key, path, input, options)
```
andrewxhill, line 623: maybe it should just run this method every time.

andrewxhill: as in, even when not isNode

carsonfarmer: Yeh, I don't think there's anything actually node specific happening here, so we could just run this way every time.

PR #777, opened 2020-12-14
ES Modules for all!

### LINKED ISSUE #776: ES Module builds for all @textile packages
We are updating all release processes to include ES modules and CJS builds. Time to make this official.

carsonfarmer:
## Description

This PR sets us up to build and release ESM and CJS outputs for external consumption. It is the equivalent to https://github.com/textileio/js-threads/pull/633 over in js-threads.

Fixes #776 

## Type of change

## How Has This Been Tested?

No new tests required. In fact, a good test that this is working correctly is that no existing tests break! We should create some tests for ES module imports though.

## Checklist:


packages/crypto/package.json
```diff
@@ -1,32 +1,33 @@
...
-	"devDependencies": {
-		"libp2p-crypto": "^0.17.8"
-	},
-	"dependencies": {
-		"@types/ed2curve": "^0.2.2",
-		"ed2curve": "^0.3.0",
-		"multibase": "^3.1.0",
-		"tweetnacl": "^1.0.3"
-	}
+  "name": "@textile/crypto",
```
carsonfarmer, line 2: was using tabs for some reason?!

.gitignore
```diff
@@ -1,7 +1,7 @@
 # Untracked docs changes
 docs
 compiled-docs
-tsdoc-metadata.json
+**/tsdoc-metadata.json
```
carsonfarmer, line 4: according to those files, we're not supposed to commit them to vc

package.json
```diff
@@ -7,7 +7,9 @@
     "clean": "lerna clean && lerna run clean",
     "bootstrap": "lerna bootstrap",
     "publish": "npm run build && lerna publish",
-    "build": "lerna run build",
+    "build": "npm run build:cjs && npm run build:esm",
```
carsonfarmer, line 10: note that there is no prepublishOnly scripts here... so it is "manual", though we do have build && lerna publish to smooth this over for us

package.json
```diff
@@ -7,7 +7,9 @@
...
+    "build:cjs": "lerna run build --stream",
+    "build:esm": "rollup -c",
```
carsonfarmer, line 12: rollup script runs at the top level. this makes for a pretty fast build process, and makes sure that things are built in the right order. it doesn't do type-checker, so we leave that to our existing cjs build process, which is why that one should always run first.

package.json
```diff
@@ -46,13 +48,20 @@
   "license": "MIT",
   "dependencies": {},
   "devDependencies": {
+    "@lerna/batch-packages": "^3.16.0",
```
carsonfarmer, line 51: new deps...

packages/buckets/package.json
```diff
@@ -2,8 +2,9 @@
   "name": "@textile/buckets",
   "version": "4.1.0",
   "description": "Web-gRPC wrapper client for managing Textile Buckets.",
-  "main": "dist/index",
-  "types": "dist/index",
+  "main": "dist/cjs/index.js",
```
carsonfarmer, line 5: all package.json files have been updated with this. 1) more explicit path to specific files, 2) types only referenced in cjs path, though they are also build (as a single file) in the esm folder. module refers to dist/esm/* so that builders/packagers know where to look. in node, if someone is using es modules, it'll link to esm, otherwise, main. main is always the fallback, so older build systems should remain unaffected

packages/buckets/package.json
```diff
@@ -13,7 +14,7 @@
   },
   "scripts": {
     "build": "npx tsc -b ./tsconfig.json",
-    "clean": "npx rimraf ./dist ./tsconfig.tsbuildinfo",
+    "clean": "npx rimraf ./dist",
```
carsonfarmer, line 17: the whole build setup is now in dist, so we can just delete that

carsonfarmer: you _might_ want to delete these build assets before pulling these changes so you don't have these hanging around?

rollup.config.js
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+import { getPackages } from '@lerna/project'
```
carsonfarmer, line 2: the full rollup config. we are essentially rolling up all vendor deps and _this_ (reach package in turn) packing into a single file, and then externally importing all other @textile packages. if we find that some of our vendor deps also start supporting es modules, then we can add them to the external array here.

the benefit is that browsers and other env can in theory tree-shake and lazily load these a bit better. ESM is the future, so we're setup for that now

rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+import filterPackages from '@lerna/filter-packages'
+import batchPackages from '@lerna/batch-packages'
+import typescript from '@wessberg/rollup-plugin-ts'
+import commonjs from '@rollup/plugin-commonjs'
+
+import resolve from '@rollup/plugin-node-resolve'
```
carsonfarmer, line 8: we have nodejs, json, typescript, etc deps, so we have plugins for all that stuff

rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+import json from '@rollup/plugin-json'
+import path from 'path'
+import minimist from 'minimist'
+
+const external = [
+  // Threads
+  '@textile/threads-client',
```
carsonfarmer, line 15: all textile stuff in here, plus other es modules as we notice them

rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+  const packages = await getPackages(__dirname)
+  const filtered = filterPackages(packages, scope, ignore, false, true)
+
+  return batchPackages(filtered).reduce((arr, batch) => arr.concat(batch), [])
+}
+
+/**
+ * @returns Promise<RollupConfig>
+ */
+async function main() {
```
carsonfarmer, line 53: this is the main config setup. we can add other build types of iife and umd etc.

carsonfarmer (commented):
Here's the full es module build PR for ya @andrewxhill !

asutula (approved):

PR #781, opened 2020-12-15
Start using conventional commits auto-publishing

carsonfarmer:

## Description

This switching to using Lerna's built-in support for conventional commits handling. This will read our commit messages (so even if you squash, you should still use conventional commit messages) to auto-generate our changelogs, and these are automatically committed and pushed on releases. This will also take care of automatically determining sermver bumps if we let it/want it to. Including managing pre-releases and more. We can always override these settings, so we're not locked in, but it does alleviate a lot of the cognitive burden of trying to remember when to bump on major/minor etc.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Should not affect tests

## Checklist:

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -6,6 +6,9 @@
   "command": {
     "version": {
       "message": "chore: release/publish"
+    },
+    "publish": {
+      "conventionalCommits": true
     }
   }
 }


asutula (approved):

asutula:
I see commitsar is still disabled here (which is great)... but did you enable some pre commit hook like husky to enforce any conventional commit style?

carsonfarmer:
Yup :)

carsonfarmer:
It'll require a fresh `npm i` locally to make sure you have all the deps (like husky), but after that, you're commits should be linted locally.

PR #786, opened 2020-12-17
listDBs/listThreads api changes

carsonfarmer:

## Description

These are proposed/possible API updates for listDBs and listThreads on the hub. This would require a minor change on the Threads API as well (to not return the extra info). We could _also_ just not make the change on the Threads API, and include that information here, but we'd need to make some additional API calls on the Hub to 'getDBInfo'.

## Type of change

## How Has This Been Tested?

New tests should be created to ensure the right behavior with and without the hub headers involved.

## Checklist:


packages/hub-threads-client/src/client.ts
```diff
@@ -16,8 +16,7 @@ const logger = log.getLogger('users')
 
 interface GetThreadResponse {
   id: string
```
carsonfarmer, line 18: So listThreads no longer returns if a given thread is a db or not.

packages/hub-threads-client/src/client.ts
```diff
@@ -50,7 +49,6 @@ Client.prototype.getThread = async function (
             if (message) {
               const res = {
                 name: message.getName(),
-                isDb: message.getIsDb(),
```
carsonfarmer, line 53: Implementation change for the above interface/return type change.

packages/hub-threads-client/src/client.ts
```diff
@@ -66,6 +64,33 @@ Client.prototype.getThread = async function (
...
 }
 
+// Private listDBs method
+const oldListDBs = Client.prototype.listDBs
+
+/**
+ * Lists all known DBs.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
```
carsonfarmer, line 74: Note the `ts-ignore`, because this change is interface-breaking re: upstream `@textile/threads-client` `listDBs`.

packages/hub-threads-client/src/client.ts
```diff
@@ -66,6 +64,33 @@ Client.prototype.getThread = async function (
...
+Client.prototype.listDBs = async function (
+  ctx?: Context,
+): Promise<Record<string, GetThreadResponse | undefined>> {
```
carsonfarmer, line 77: Proposal, this just "lists" (we could actually use a list) the threads which are dbs. It does _not_ include the key info or addrs. To get this information, a second call to getDBInfo would be required. This seems useful, because often you just want to list DBs, but you only need the extra info if you are going to do something with a given thread. 

Currently in threads-client, this information (addrs and key) are included by default, but since this isn't readily available here, I'm suggesting we drop this... perhaps leaving it backward compatible from threads-client?

packages/hub-threads-client/src/client.ts
```diff
@@ -66,6 +64,33 @@ Client.prototype.getThread = async function (
...
+    // We're probably on the Hub
+    const threads = await this.listThreads(ctx)
+    for (const thread of threads) {
+      dbs[thread.id] = thread
+    }
+    return dbs
+  }
+  const threads = await oldListDBs.bind(this)()
+  for (const [id, db] of Object.entries(threads)) {
+    dbs[id] = { id, name: db?.name }
```
carsonfarmer, line 89: using a record/object here, but could also just be a list/array.

carsonfarmer (commented):
Putting together simple "wrapper" implementation proposal for @sanderpick and @andrewxhill to review.

andrewxhill (approved):
looks good. the list proposal makes sense i think

packages/hub-threads-client/src/client.ts
```diff
@@ -66,6 +64,33 @@ Client.prototype.getThread = async function (
...
+/**
+ * Lists all known DBs.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+Client.prototype.listDBs = async function (
+  ctx?: Context,
+): Promise<Record<string, GetThreadResponse | undefined>> {
+  const dbs: Record<string, GetThreadResponse | undefined> = {}
+  if (ctx?.get('x-textile-api-sig')) {
```
andrewxhill, line 79: the _only_ thing i can think... if they call this before the sig has been created will they get a helpful error or will it say the api doesn't exist?

sanderpick: Cool! This made me realize... since the dashboard code will in some way need to list threads, etc., it probably makes sense to utilize this code. We'll just have to add handling for normal dev/org sessions (like the CLI), which can be stored in a browser cookie / local storage.

packages/hub-threads-client/src/client.ts
sanderpick: Nice, makes sense... we can drop from the go-threads API too.

sanderpick (approved):
😎

packages/hub-threads-client/src/client.ts
carsonfarmer: no actually I'm wrong, they'll get the right error message... that they haven't authenticated or whatever

andrewxhill: oh, i thought it would see it missing so try to call the other api. which wouldn't exist

carsonfarmer: yup, that's right... I'm crazy. added "catch" for this situation.

PR #814, opened 2021-01-06
Hub admin client/APIs

### LINKED ISSUE #793: Stick with cids rather than multiformats/cids for now
The change is going to be great, but for now, let's stick with the more "stable" lib.

### LINKED ISSUE #813: Implement Hub Admin APIs

carsonfarmer:
## Description

This adds pretty much full coverage of the hub admin APIs as a new sub-package. It is NOT linked or exported from the usual @textile/hub meta-package on purpose, because this is mostly an internal use package.

Note that I have also tagged this PR as under "discussion". I went with some generic names etc that made sense to me. Additionally, the APIs reflect how I think they would be consumed by a UI-driven app, but I'd like to have @brunellaa's opinion here if possible. Additionally, since this package will be marked internal only, I don't mind making future breaking changes if we have to.

Fixes #813
Fixes #793 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

## How Has This Been Tested?

All existing tests continue to pass, and a whole suite of new tests have been added.

TODO:
- Currently we are missing tests for the billing apis, its' not clear to me if we even should at this API level anyway?

## Checklist:


etc/hub.api.md
```diff
@@ -750,7 +750,9 @@ export interface Period {
 }
 
 // @public
-export type Private = Identity;
+export interface Private extends Identity {
+    bytes: Uint8Array;
```
carsonfarmer, line 754: this is a breaking interface change... there doesn't appear to be any consequences in our own codebase, but it should lead to a minor bump (we're adding a value). Check me on this though pls.

packages/admin/README.md
```diff
@@ -0,0 +1,20 @@
+# @textile/hub-admin
+
+**Info**
+
+This is not a supported developer library.
```
carsonfarmer, line 5: hoping this is enough to dissuade would be installers from using this :P

packages/admin/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@textile/hub-admin",
```
carsonfarmer, line 2: calling this hub-admin... open to discussion!

packages/admin/src/admin.spec.ts
```diff
@@ -0,0 +1,408 @@
+import { Context } from '@textile/context'
```
carsonfarmer, line 1: The main tests, mostly ported from the go tests, with some js stuff in there.

packages/admin/src/admin.ts
```diff
@@ -0,0 +1,197 @@
+import { Context } from '@textile/context'
+import {
+  CopyAuthOptions,
+  GrpcAuthentication,
+  WithUserAuthOptions,
+} from '@textile/grpc-authentication'
+import * as api from './api'
+
+export class Admin extends GrpcAuthentication {
```
carsonfarmer, line 9: The actual admin class, following @andrewxhill 's pattern for this. The name is up for debate!

packages/admin/src/api/index.ts
```diff
@@ -0,0 +1,516 @@
+import type { ContextInterface } from '@textile/context'
```
carsonfarmer, line 1: The actual API implementations, after @andrewxhill's pattern. I'm being pretty careful here about returning "nice" things to the end user. So make sure that all feels right in your review. I'm using strings for keys, because that's how we'll consume them elsewhere... but we could just as easily return the actual key classes (PrivateKey, PublicKey)...

packages/admin/src/api/index.ts
```diff
@@ -0,0 +1,516 @@
...
+  api: GrpcConnection,
+  ctx?: ContextInterface,
+): Promise<void> {
+  logger.debug('signout request')
+  const req = new pb.SignoutRequest()
+  await api.unary(APIService.Signout, req, ctx)
+  return
+}
+
+export interface SessionInfoResponse {
```
carsonfarmer, line 83: Each return type here is grouped and placed above the function that returns it..

packages/admin/src/api/index.ts
```diff
@@ -0,0 +1,516 @@
...
+  const req = new pb.GetIdentityRequest()
+  const res: pb.GetIdentityResponse = await api.unary(
+    APIService.GetIdentity,
+    req,
+    ctx,
+  )
+  return privateKeyBytesToString(res.getIdentity_asU8())
+}
+
+export enum KeyType {
```
carsonfarmer, line 133: Using an enum here... do we use key type like this anywhere else?

packages/admin/src/api/index.ts
```diff
@@ -0,0 +1,516 @@
...
+export interface OrgInfo {
+  key: string
+  name: string
+  slug: string
+  host: string
+  members: Array<OrgMember>
+  createdAt: number
+}
+
+const pbToOrgInfo = (orgInfo?: pb.OrgInfo): OrgInfo => {
```
carsonfarmer, line 225: We use a pbTo* function here so that we can properly parse the keys etc.

packages/admin/src/api/index.ts
```diff
@@ -0,0 +1,516 @@
...
+  return
+}
+
+/**
+ * (Re-)enables billing for an account, enabling usage beyond the free quotas.
+ * @param ctx Context containing gRPC headers and settings.
+ * These will be merged with any internal ctx.
+ * @internal
+ */
+export async function setupBilling(
```
carsonfarmer, line 364: I have these here, but I'm not testing them right now.

packages/admin/src/api/index.ts
```diff
@@ -0,0 +1,516 @@
...
+  period?: Period
+}
+
+export interface Period {
+  unixStart: number
+  unixEnd: number
+}
+
+export interface Customer {
+  // TODO: Have to properly unmarshall this object rather than use toObject
```
carsonfarmer, line 410: This todo might be relevant when we actually start using these apis and creating tests for them. toObject returns base64 encoded strings, which is usually not what we want.

packages/admin/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './admin'
```
carsonfarmer, line 1: that's it, that's all there is... @andrewxhill lemme know if this * thing isn't nice for docs etc.

packages/admin/src/spec.utils.ts
```diff
@@ -0,0 +1,76 @@
+import axios from 'axios'
```
carsonfarmer, line 1: Since most of the other test suites have something like this... but it actually wraps functionality for this package, we could move to having them import this stuff from here, reducing copy pasta errors in the future. Can create an issue for this.

packages/admin/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": { "outDir": "./dist/cjs", "rootDir": "./src" },
+  "references": [
+    { "path": "../grpc-authentication" },
```
carsonfarmer, line 5: auto-generated

packages/buckets/package.json
```diff
@@ -32,12 +32,12 @@
     "@textile/security": "^0.7.0",
     "@textile/threads-id": "^0.4.0",
     "abort-controller": "^3.0.0",
+    "cids": "^1.1.4",
     "event-iterator": "^2.0.0",
-    "loglevel": "^1.6.8",
-    "multiformats": "^4.4.1"
```
carsonfarmer, line 37: so long!

packages/crypto/src/identity.ts
```diff
@@ -26,7 +26,12 @@ export interface Identity {
 /**
  * Private is an alias for the {@link Identity} interface.
  */
-export type Private = Identity
+export interface Private extends Identity {
+  /**
+   * Return the protobuf-encoded bytes of this private key.
+   */
+  bytes: Uint8Array
```
carsonfarmer, line 33: breaking change

packages/crypto/src/index.ts
```diff
@@ -5,7 +5,9 @@ export {
   decrypt,
   encrypt,
   extractPublicKeyBytes,
+  privateKeyBytesToString,
```
carsonfarmer, line 8: breaking change, but only in that we're adding stuff

packages/crypto/src/keypair.ts
```diff
@@ -1,24 +1,15 @@
...
 import {
   decrypt,
   encrypt,
   privateKeyFromString,
+  privateKeyToString,
   publicKeyBytesFromString,
   publicKeyToString,
 } from './utils'
 
-/**
```
carsonfarmer, line 13: moved this, so we're still exporting this from the package

packages/users/src/users.ts
```diff
@@ -13,7 +13,6 @@ import {
 } from '@textile/grpc-authentication'
 import { GetThreadResponse } from '@textile/hub-threads-client'
 import { KeyInfo, UserAuth } from '@textile/security'
-import log from 'loglevel'
```
carsonfarmer, line 16: not really needed here

carsonfarmer (commented):
Some comments for your reviews @andrewxhill and @brunellaa

packages/admin/README.md
andrewxhill: maybe the inverse

packages/admin/src/admin.ts
andrewxhill: nice! might as well i guess, though the benefit was mostly for getting nice docs, which wouldn't matter here.

packages/admin/src/index.ts
andrewxhill: well.. doesn't matter right? since we wouldn't update docs

andrewxhill (approved):

packages/admin/README.md
carsonfarmer: Example?

packages/admin/src/admin.ts
carsonfarmer: Yeh, it was just a nice pattern to use because it actually makes the functions themselves a lot more reusable.

packages/admin/src/index.ts
carsonfarmer: good point

packages/admin/README.md
andrewxhill: oh, ha, i just meant it may encourage use :D

carsonfarmer: lol, I see! yes that's entirely possible :P

Issue #757, opened 2020-12-08
Migrate threads-client into monorepo


andrewxhill:
Other packages TBD as convenient.

carsonfarmer:
Done with current master

PR #820, opened 2021-01-07
Move all utility packages from js-threads

### LINKED ISSUE #817: Mirgate: Move @textile/grpc-transport over here

carsonfarmer:
## Description

This is the precursor to moving everything over from js-threads into this mono-repo. This includes all utility libraries. It does not include threads-client (see #757), which currently requires its own testing env that doesn't quite match what we have here.

Note that this PR includes a ton of commits, to preserve (well, mostly) the commit history of all packages/updates in js-threads.

Fixes #817 

## Type of change

Administrative

## How Has This Been Tested?

Should not affect tests

## Checklist:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Typescript/Javascript libs for interacting with Textile APIs.",
   "main": "packages/hub",
   "scripts": {
-    "clean": "lerna clean && lerna run clean",
+    "clean": "npx rimraf dist && lerna clean --yes && lerna run clean",
```
carsonfarmer, line 7: faster cleanup and remove this "extra" dist at the top level that is created by the esm build step

packages/admin/tsconfig.json
```diff
@@ -1,8 +1,21 @@
...
   "extends": "../../tsconfig.json",
-  "compilerOptions": { "outDir": "./dist/cjs", "rootDir": "./src" },
+  "compilerOptions": {
+    "outDir": "./dist/cjs",
+    "rootDir": "./src"
+  },
   "references": [
-    { "path": "../grpc-authentication" },
-    { "path": "../grpc-connection" }
+    {
```
carsonfarmer, line 8: all sub-packages have now been updated with their references, so all local references should be up to date

packages/context/README.md
```diff
@@ -0,0 +1,19 @@
+# @textile/context
```
carsonfarmer, line 1: I've switching all incoming readmes to use the "pattern". but note, we should decide if @textile/hub should still be the main entry point for everything, or if it should be something like @textile/textile?

carsonfarmer: Good discussion point @andrewxhill

packages/context/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "access": "public"
+  },
+  "license": "MIT",
+  "scripts": {
+    "build": "npm run build:cjs",
+    "build:cjs": "npx tsc -b ./tsconfig.json",
+    "clean": "npx rimraf ./dist",
+    "test:browser": "polendina --cleanup src/**/*.spec.ts --webpack-config ../../webpack.test.js"
+  },
+  "repository": "github:textileio/js-threads",
```
carsonfarmer, line 21: these will have to be updated

carsonfarmer (commented):
A lot of administrative changes here. You can probably ignore most. See my comments where there is some room for discussion. For instance, I haven't moved for @textile/threads or @textile/threads-client yet...

packages/context/README.md
andrewxhill: what would textile/textile be?

andrewxhill: {textile}/hub <- textile already is in the namespace. i think all packages should be designed to work with the hub. you can use them in a mode that wont touch the hub, but that's an option the developer chooses. they are still hub packages none-the-less.

andrewxhill (approved):
this is great. i think this might make our lives a bit easier every week!

packages/context/README.md
carsonfarmer: yup, makes sense. @textile/hub it is!

carsonfarmer:
Note that I am _not_ going to squash this before merging in order to preserve the long history from js-threads.

PR #842, opened 2021-01-21
feat(package): adds start of buck push util

andrewxhill:


packages/buckets-node-sync/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist/cjs",
+    "rootDir": "./src"
+  },
+  "references": [
```
carsonfarmer, line 7: Can you run `npm run references` to make sure this is up to date. I don't need buckets in here, but I do see hub-threads-client, which this package doesn't reference directly.

packages/buckets-node-sync/src/cli.ts
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env node
+
+import yargs, {Argv, string} from "yargs"
+import { execute, RunOutput } from './index'
+
+// Example:
+// buckjs push --api=http://127.0.0.1:3007 -k blp5ihqcs5zdryhc23fx2x7l5ru -s b6dtmohcxvcd3uqnbpx6b7ygb6m7hu4cslgsoroy -t bafku6ozemzot6fwazeolxag2kvd3c753nudnwat4mfoinatiffeicdi -n test -p /Users/andrewhill/Textile/textile/js-textile/packages/buckets-node-sync/test/website
```
carsonfarmer, line 7: So cool!

carsonfarmer (changes requested):
Just one minor update required here @andrewxhill, otherwise, looks sweet!

packages/buckets-node-sync/src/cli.ts
andrewxhill: just pushed an update that cleans it all up. allows same ENV var pattern etc (see readme)

andrewxhill: Screen Shot 2021-01-22 at 10 15 49 AM

carsonfarmer (approved):

packages/buckets-node-sync/tsconfig.json
andrewxhill: done

packages/buckets-node-sync/src/cli.ts
andrewxhill: those aren't really keys/secrets, don't worry :D

PR #862, opened 2021-01-26
fix(api): browser enabled pullpath

andrewxhill:


packages/buckets/src/buckets.spec.ts
```diff
@@ -210,27 +215,41 @@ describe('Buckets...', function () {
...
       const chunks = client.pullPath(rootKey, 'dir1/file1.jpg', {
         progress: (num) => (length = num || 0),
       })
-      const pth = path.join(__dirname, '../../..', 'testdata')
-      const stream = fs.createWriteStream(path.join(pth, 'output.jpg'))
-      // Should be an AsyncIterable
-      for await (const chunk of chunks) {
-        stream.write(chunk)
+      if (isBrowser) {
+        let result = new Uint8Array();
```
andrewxhill, line 226: uploaded file is different so has a different set of expectations

packages/buckets/src/buckets.spec.ts
```diff
@@ -159,7 +164,7 @@ describe('Buckets...', function () {
       const res = await client.pushPath(rootKey, 'dir1/file1.jpg', file, {
         progress: (num) => (length = num || 0),
       })
-      expect(length).to.equal(54)
+      expect(length).to.equal(620047)
```
andrewxhill, line 167: used a much larger test input string to ensure we span more chunks

packages/buckets/src/buckets.spec.ts
```diff
@@ -210,27 +215,41 @@ describe('Buckets...', function () {
...
         progress: (num) => (length = num || 0),
       })
-      const pth = path.join(__dirname, '../../..', 'testdata')
-      const stream = fs.createWriteStream(path.join(pth, 'output.jpg'))
-      // Should be an AsyncIterable
-      for await (const chunk of chunks) {
-        stream.write(chunk)
+      if (isBrowser) {
+        let result = new Uint8Array();
+        const append = (target: Uint8Array, addition: Uint8Array) => {
```
carsonfarmer, line 227: If we don't have this in a util module somewhere, we probably should. I'd guess we use this pattern elsewhere.

carsonfarmer (approved):
Yeh looks good @andrewxhill , nice to have some more browser-specific tests in there for sure.

packages/buckets/src/buckets.spec.ts
andrewxhill: with all the packages moving together, it seems like a nice enhancement in the near future will be:

1. a unified set of utils that users might want across the board
2. a unified set of utils used in tests all over the place

PR #866, opened 2021-01-27
Starting to implement push paths

carsonfarmer:
## Description

Implements the bulk upload API on Buckets. This includes some nice idiomatic JS features for creating and consuming streams. The code is lazily evaluated at the time of consumption, so you can setup a whole array of pushes ahead of time and only issue them when required. There is no need for a progress callback, because the consumer can simply respond to the updates and do their progress stuff as needed.

Fixes # (issue)

## Type of change

## Checklist:


packages/buckets/src/buckets.ts
```diff
@@ -660,7 +662,15 @@ export class Buckets extends GrpcAuthentication {
     input: any,
     options?: PushOptions,
   ): Promise<PushPathResult> {
-    return bucketsPushPathNode(this, key, path, input, options)
+    return bucketsPushPath(this, key, path, input, options)
```
carsonfarmer, line 665: Dropping the "node" version in favor of just the one, which _is_ actually the "node" style emitter version.

packages/admin/package.json
```diff
@@ -25,7 +25,7 @@
     "@textile/grpc-authentication": "^3.2.0",
     "@textile/grpc-connection": "^2.3.0",
     "@textile/grpc-transport": "^0.3.0",
-    "@textile/hub-grpc": "^2.5.0",
+    "@textile/hub-grpc": "^2.6.0",
```
carsonfarmer, line 28: These are all getting bumped.

packages/buck-util/tsconfig.json
```diff
@@ -1,11 +1,15 @@
...
+    "outDir": "./dist/cjs",
+    "rootDir": "./src"
+  },
   "references": [
-    { "path": "../buckets" },
-    { "path": "../context" },
-    { "path": "../grpc-connection" },
-    { "path": "../security" },
-    { "path": "../transport" }
+    {
```
carsonfarmer, line 8: Updating these using builtin tooling and then linter

packages/buckets/src/api/index.ts
```diff
@@ -1,4 +1,5 @@
 import { grpc } from '@improbable-eng/grpc-web'
+import { Repeater } from '@repeaterjs/repeater'
```
carsonfarmer, line 2: This is a _much_ nicer lib for dealing with "pushable" async iterables. It is cleaner, provides lazy evaluation, and the API feels much more like the Promise-style API for creating new promises.

packages/buckets/src/api/index.ts
```diff
@@ -46,16 +49,17 @@ import {
...
   Status,
 } from '@textile/buckets-grpc/api/bucketsd/pb/bucketsd_pb_service'
 import { Context, ContextInterface } from '@textile/context'
 import { GrpcConnection } from '@textile/grpc-connection'
 import { WebsocketTransport } from '@textile/grpc-transport'
 import CID from 'cids'
-import { EventIterator } from 'event-iterator'
+import drain from 'it-drain'
 import log from 'loglevel'
+// @ts-expect-error
```
carsonfarmer, line 61: This doesn't have types, hence the error expectation here.

packages/buckets/src/api/index.ts
```diff
@@ -73,12 +77,15 @@ import {
...
   PushOptions,
   PushPathResult,
+  PushPathsResult,
   RemovePathOptions,
   RemovePathResponse,
   Root,
 } from '../types'
 import { File, normaliseInput } from './normalize'
 
+export { File }
```
carsonfarmer, line 87: Forgot to export this in the past, useful to have this.

packages/buckets/src/api/index.ts
```diff
@@ -426,112 +432,6 @@ export async function bucketsPushPath(
   return new Promise<PushPathResult>(async (resolve, reject) => {
     // Only process the first input if there are more than one
     const source: File | undefined = (await normaliseInput(input).next()).value
-    const client = grpc.client<
```
carsonfarmer, line 429: This mess is basically due to removing the old function named bucketsPushPath, and replacing it with buckesPushPathNode... nothing really new here.

packages/buckets/src/api/index.ts
```diff
@@ -589,14 +489,14 @@ export async function bucketsPushPathNode(
       if (status && status.code !== grpc.Code.OK) {
         return reject(new Error(status.details))
       } else {
-        return resolve()
+        return reject(new Error('undefined result'))
```
carsonfarmer, line 492: Fixes some ts issues, and is actually the correct action.

packages/buckets/src/api/index.ts
```diff
@@ -631,8 +531,114 @@ export async function bucketsPushPathNode(
...
+ * @internal
+ */
+export function bucketsPushPaths(
+  api: GrpcConnection,
+  key: string,
+  input: any,
+  opts?: Omit<PushOptions, 'progress'>,
+  ctx?: ContextInterface,
+): AsyncIterableIterator<PushPathsResult> {
+  return new Repeater<PushPathsResult>(async (push, stop) => {
```
carsonfarmer, line 549: Simply returns an AsyncIterableIterator that only kicks off once you actually pull from it. This is nice, and makes it feel like a parallel push for an iterable of files. This function will take anything that can be converted to an async iterable of files... which includes iterables, async iterables, streams, etc.

packages/buckets/src/api/index.ts
```diff
@@ -631,8 +531,114 @@ export async function bucketsPushPathNode(
...
+        return stop(AbortError)
+      }
+      const obj: PushPathsResult = {
+        path: message.getPath(),
+        root: fromPbRootObjectNullable(message.getRoot()),
+        cid: new CID(message.getCid()),
+        pinned: message.getPinned(),
+        size: message.getSize(),
+      }
+      push(obj)
```
carsonfarmer, line 579: Just push the object onto the iterable for consumers. We don't need a progress callback, because they can simply respond to progress within the async iterable loop.

packages/buckets/src/api/index.ts
```diff
@@ -631,8 +531,114 @@ export async function bucketsPushPathNode(
...
+          // Naively chunk into chunks smaller than CHUNK_SIZE bytes
+          for (const chunklet of genChunks(data as Uint8Array, CHUNK_SIZE)) {
+            chunk.setData(chunklet)
+            req.setChunk(chunk)
+            stream.write(req)
+          }
+        }
+      }
+      // Close out the file
+      const final = new PushPathsRequest.Chunk()
```
carsonfarmer, line 628: We need to send empty bytes for the last chunk to close out the file.

packages/buckets/src/api/index.ts
```diff
@@ -631,8 +531,114 @@ export async function bucketsPushPathNode(
...
+      final.setPath(path)
+      req.setChunk(final)
+      stream.write(req)
+    }
+
+    // We don't care about the top level order, progress is labeled by path
+    await drain(paramap(normaliseInput(input), mapper, { ordered: false }))
+    stream.end()
```
carsonfarmer, line 636: Don't close the stream until _after_ we've consumed all files (in parallel).

packages/buckets/src/api/index.ts
```diff
@@ -664,12 +670,12 @@ export function bucketsPullPath(
   opts?: { progress?: (num?: number) => void },
   ctx?: ContextInterface,
 ): AsyncIterableIterator<Uint8Array> {
-  const metadata = { ...api.context.toJSON(), ...ctx?.toJSON() }
-  const request = new PullPathRequest()
-  request.setKey(key)
-  request.setPath(path)
-  let written = 0
-  const events = new EventIterator<Uint8Array>(({ push, stop, fail }) => {
+  return new Repeater<Uint8Array>((push, stop) => {
```
carsonfarmer, line 673: Switching to the Repeater pattern here, which shouldn't affect any external APIs, but it doesn't simplify our code, and should be a more stable approach.

packages/buckets/src/api/index.ts
```diff
@@ -690,21 +696,14 @@ export function bucketsPullPath(
...
-          fail(new Error(message))
+          stop(new Error(message))
         }
         stop()
       },
     })
-    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-    return () => resp.close()
+    // Cleanup afterwards
+    stop.then(() => resp.close())
```
carsonfarmer, line 705: This ensures the stream is closed when we finish the async iterable loop.

packages/buckets/src/api/index.ts
```diff
@@ -745,21 +744,13 @@ export function bucketsPullIpfsPath(
         // _trailers: grpc.Metadata,
       ) => {
         if (status !== grpc.Code.OK) {
-          fail(new Error(message))
+          stop(new Error(message))
```
carsonfarmer, line 747: There is no fail, only stop.

packages/buckets/src/api/index.ts
```diff
@@ -996,8 +987,8 @@ export class BucketsGrpcClient {
     R extends grpc.ProtobufMessage,
     T extends grpc.ProtobufMessage,
     M extends grpc.UnaryMethodDefinition<R, T>
-  >(methodDescriptor: M, req: R, ctx?: ContextInterface): Promise<T> {
-    return new Promise<T>((resolve, reject) => {
+  >(methodDescriptor: M, req: R, ctx?: ContextInterface): Promise<T | void> {
+    return new Promise<T | void>((resolve, reject) => {
```
carsonfarmer, line 991: Nicer type checking

carsonfarmer (commented):
Here are some comments for @sanderpick and @andrewxhill

packages/buckets/src/buckets.spec.ts
```diff
@@ -385,24 +390,86 @@ describe('Buckets...', function () {
...
+      const totals = new Map([
+        ['file1.jpg', 0],
+        ['path/to/file2.txt', 0],
+      ])
+      for await (const { path, root, size } of iter) {
+        expect(['file1.jpg', 'path/to/file2.txt']).to.include(path)
+        expect(root).to.not.be.undefined
+        totals.set(path, size)
+      }
+      // TODO: For some reason I'm getting slightly larger size from buckets
```
sanderpick, line 418: If the size compared is from the push result, I believe it will be slightly different due to the IPFS adder chunking.

sanderpick (approved):
LGTM! 💪

andrewxhill (approved):

packages/buckets/src/buckets.spec.ts
carsonfarmer: Cool, that makes sense. I've updated the test to reflect this.

PR #868, opened 2021-01-28
deduplicate all the test methods shared across packages

andrewxhill:


carsonfarmer (approved):
LOVE it. Nice one @andrewxhill

PR #894, opened 2021-02-04
Migrate threads into textile + new test setup

carsonfarmer:
Moves all of threads over to js-textile, and sets up dual test setup (hub vs "raw" threads daemons). Also makes a few updates to threads along the way. Note that I am just moving the threads code here "fresh". Because the entire git commit history from threads was previously imported into js-textile, and there were only about 2 updates to threads since then.


.github/workflows/test.yml
```diff
@@ -1,16 +1,85 @@
...
     branches:
       - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
-  test:
-    name: Test
+  threads:
```
carsonfarmer, line 11: we now have threads tests, and hub tests.

packages/id/package.json
```diff
@@ -18,11 +18,11 @@
     "clean": "npx rimraf ./dist",
     "test:browser": "polendina --cleanup src/**/*.spec.ts --webpack-config ../../webpack.test.js"
   },
-  "repository": "github:textileio/js-threads",
+  "repository": "github:textileio/js-textile",
```
andrewxhill, line 21: 🎉

andrewxhill (approved):

PR #919, opened 2021-02-08
fix(err): ensure latest root before handling deletes

andrewxhill:


carsonfarmer (approved):
Good catch. Would it be "more" useful to actually just ensure the last iter result _is_ the last one processed? I think this is doable on the buckets api side?

PR #991, opened 2021-03-16
chore(docs): adds mv, adds push paths

andrewxhill:


andrewxhill:
primary purpose was to add this, plus update the sidebar with missing methods

Screen Shot 2021-03-16 at 9 32 00 AM

asutula (approved):

PR #1207, opened 2021-10-27
Support more schema formats (doesn't affect remote)

carsonfarmer:

## Description

Supports additional schema types in the local threaddb repo. This doesn't affect the remote in any way, so incompatibilities there will cause issues for folks.


carsonfarmer:
going to just merge this to keep the release process going

